PR #83, opened 2025-10-09
AWS Karpenter

alex-hunt-materialize:
* Sets the node selector for cert-manager pods
* Configures Karpenter for managing nodes.

Don't be scared off by the 90k+ lines of the PR. Most of that is just a generated file describing AWS instance types.

The `base` node group is only there to run coredns and Karpenter. All other pods should go to nodes spawned by Karpenter in either the generic or materialize node pools.


aws/examples/simple/main.tf
```diff
@@ -325,6 +419,60 @@ locals {
     local.materialize_instance_namespace,
     local.materialize_instance_name
   )
+
+  ami_selector_terms = [{ "alias" : "bottlerocket@latest" }]
```
bobbyiliev: Are we good using `@latest` here or should we pin to a specific version?

bobbyiliev (approved):

aws/examples/simple/main.tf
alex-hunt-materialize: As this is just the example, I think latest is fine. We'd have to update it frequently to ensure we don't force an insecure version otherwise.

aws/modules/karpenter-ec2nodeclass/main.tf
```diff
@@ -0,0 +1,145 @@
+locals {
+  # Generated with:
+  # aws ec2 describe-instance-types --query "InstanceTypes[]"
+  instance_descriptions = jsondecode(file("${path.module}/instance-descriptions.json"))
```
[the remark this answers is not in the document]
alex-hunt-materialize: Currently just updated manually when needed, but I think Shiwam's suggestion might work.

PR #110, opened 2025-11-26
aks public cluster with api server vnet integration

jshiwamV:
- Public Cluster Endpoint with api server vnet integration when vnet encryption is disabled works with latest VM skus.
- Pod to ControlPlane traffic to api-server is contained inside VPC via api server vnet integration
- Endpoint is publicly accessible to restricted CIDR for external usage
- Kubernetes outbound access is configured via [nat gateway because it allows more outbound traffic flows per ip address](https://learn.microsoft.com/en-us/azure/aks/nat-gateway)

The cluster is not private though, if we make the cluster private we would need a bastion host to connect to the cluster from outside of vpc. The current configuration mimics the behaviour of public access + keeping internal aks traffic contained inside vnet. As per internal slack discussion, we don’t want fully private cluster hence the current configuration should work for us.


azure/modules/aks/main.tf
```diff
@@ -52,6 +60,15 @@ resource "azurerm_kubernetes_cluster" "aks" {
...
   }
 
+  # API Server VNet Integration - Projects API server into a delegated subnet
+  # Can be used for BOTH public and private clusters
+  # Reference: https://learn.microsoft.com/en-us/azure/aks/api-server-vnet-integration
+  api_server_access_profile {
+    virtual_network_integration_enabled = var.enable_api_server_vnet_integration
+    subnet_id                           = var.api_server_subnet_id
+    authorized_ip_ranges                = var.api_server_authorized_ip_ranges
+  }
```
jubrad: does this also need to be authorized by the CIDR of the VPC?

jshiwamV: Huh, i dont think we need to add vpc address to authorised_cidr. I went through the docs but it was still not clear, so here is how i tested that the, I removed vnet cidr from authorised_ip_range, and tried to connect to api-server from one of the running pods in the cluster

```
/ # TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
/ # wget --header "Authorization: Bearer $TOKEN" https://kubernetes.default.svc/api
Connecting to kubernetes.default.svc (20.1.0.1:443)
wget: note: TLS certificate validation not implemented
saving to 'api'
api                  100% |************************************************************************************************************************************************************|   218  0:00:00 ETA
'api' saved
/ # cat api
{
  "kind": "APIVersions",
  "versions": [
    "v1"
  ],
  "serverAddressByClientCIDRs": [
    {
      "clientCIDR": "0.0.0.0/0",
      "serverAddress": "sj-manual-aks-nkeynove.hcp.westeurope.azmk8s.io:443"
    }
  ]
}/ # 

```
it is able to connect to the server, which means the pods inside the cluster can access the api server. i think [authorized ip ranges](https://learn.microsoft.com/en-us/azure/aks/api-server-authorized-ip-ranges?tabs=azure-cli#create-an-aks-cluster-with-api-server-authorized-ip-ranges-enabled) is specifically for external access not internal.

Issue #113, opened 2025-11-27
Enhance security Group rules with least-privilege principles.


jshiwamV:
**Security Groups**:
    - Tighten security groups with access only from necessary sources
    - Implement separate security groups for control plane, worker nodes, and database access
    - Use security group rules that follow least-privilege principles

PR #119, opened 2025-12-05
network security and lb access public

jshiwamV:
- public access only possible via nlb
- users can restrict public access to their whitelisted cidrs.
- AKS subnet only takes in traffic within vpc blocking all external traffic except the traffic frm loadbalancer.

Related Issue: https://github.com/MaterializeInc/materialize-terraform-self-managed/issues/113

### DIFF
--- a/azure/modules/load_balancers/variables.tf
+++ b/azure/modules/load_balancers/variables.tf
@@ -83,9 +83,9 @@ variable "resource_id" {
 }
 
 variable "internal" {
-  description = "Whether the load balancer is internal to the VNet. Defaults to false (public) to allow external access to Materialize. Set to true for VNet-only access."
+  description = "Whether the load balancer is internal to the VNet. Defaults to true (internal) to allow internal access to Materialize. Set to false for public access."
   type        = bool
-  default     = false
+  default     = true
   nullable    = false
 }
 


azure/modules/load_balancers/variables.tf
```diff
@@ -17,9 +17,9 @@ variable "resource_id" {
 }
 
 variable "internal" {
-  description = "Whether the load balancer is internal to the VPC."
+  description = "Whether the load balancer is internal to the VNet. Defaults to false (public) to allow external access to Materialize. Set to true for VNet-only access."
   type        = bool
-  default     = true
+  default     = false
   nullable    = false
 }
```
jubrad: We should probably default internal to true, it should be opt in.

jubrad (approved):
LG but we should default to internal I think.

azure/modules/load_balancers/variables.tf
```diff
@@ -44,8 +81,8 @@ variable "materialize_balancerd_https_port" {
...
-variable "ingress_cidr_blocks" {
-  description = "CIDR blocks that are allowed to reach the Azure LoadBalancers."
-  type        = list(string)
-  nullable    = false
-}
+# variable "ingress_cidr_blocks" {
+#   description = "CIDR blocks that are allowed to reach the Azure LoadBalancers."
+#   type        = list(string)
+#   nullable    = false
+# }
```
jubrad: Assuming this was also accidentally commited

jshiwamV: fixed

alex-hunt-materialize (approved):

PR #128, opened 2025-12-16
Disable TTL cache in coredns AWS

jshiwamV:
- added TTL config for aws.
- TTL 0 for kubernetes plugin in coredns, now disables the cache for kubernetes records.

### DIFF
--- a/aws/modules/eks/main.tf
+++ b/aws/modules/eks/main.tf
@@ -48,8 +48,6 @@ module "eks" {
           }
         }
         corefile = <<-EOT
-          # This is the stock EKS configuration for CoreDNS, with modifications
-          # noted inline.
           .:53 {
               errors
               health {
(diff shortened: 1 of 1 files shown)


alex-hunt-materialize (dismissed):

aws/modules/eks/main.tf
```diff
@@ -15,6 +15,73 @@ module "eks" {
...
+                  lameduck 5s
+              }
+              ready
+              kubernetes cluster.local in-addr.arpa ip6.arpa {
+                  ttl 0
+                  pods insecure
+                  fallthrough in-addr.arpa ip6.arpa
+              }
+              prometheus :9153
+              cache 30
```
jubrad: I think we should set the following here as well.. I noticed better stability
```
                            cache 30 {
                                disable success cluster.local
                                disable denial  cluster.local
                            }
```

PR #165, opened 2026-03-12
coredns: fix spurious replacements from ephemeral access tokens

pH14:
Human note: this is a real problem I'm seeing while pointing to this module (https://github.com/MaterializeInc/mz-context-graph/actions/runs/22984168372/job/66730746089), however the entire change is Claude driven

## Summary

The kube-dns scale `terraform_data` resources use `triggers_replace` with `var.kubeconfig_data`, which embeds the GKE/EKS access token. Since the token changes on every apply (~1h lifetime), Terraform sees a trigger change on every run, attempts to replace the resource, and the destroy provisioner fails with `Unauthorized` because it runs with the stale token from state.

- Add a `cluster_identifier` variable that callers can set to a stable value (e.g. the cluster endpoint). When set, only actual cluster changes trigger re-execution of the scale commands. Defaults to `kubeconfig_data` for backwards compatibility.
- Change destroy provisioners from `on_failure = fail` to `on_failure = continue`. The scale-up is a safety net for cleanup — if the cluster is also being destroyed or the token is stale, blocking the entire apply is worse than a warning.

### Usage

```hcl
module "coredns" {
  source = "github.com/MaterializeInc/materialize-terraform-self-managed//kubernetes/modules/coredns"

  kubeconfig_data    = local.kubeconfig_data
  cluster_identifier = module.gke.cluster_endpoint  # stable, only changes on cluster recreation
  # ...
}
```

Without `cluster_identifier`, the module behaves exactly as before (triggers on the full kubeconfig).

### DIFF
--- a/kubernetes/modules/coredns/main.tf
+++ b/kubernetes/modules/coredns/main.tf
@@ -7,6 +7,10 @@ locals {
     "provisioned-by" = "materialize"
   }
 
+  # Stable trigger for kube-dns scale commands. Re-runs only when the cluster
+  # itself changes (e.g. recreation), not on every apply due to token rotation.
+  scale_trigger = coalesce(var.cluster_identifier, var.kubeconfig_data)
+
   # Corefile with TTL 0 first in the kubernetes plugin block (required for correct parsing)
   corefile = <<-EOF
     .:53 {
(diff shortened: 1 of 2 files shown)


bobbyiliev (approved):
Looks good!

PR #167, opened 2026-03-12
Add Azure enterprise example with Ory Kratos and Hydra for OIDC/SAML auth DEP-140

bobbyiliev:
Should be good for review.

- Azure enterprise example with Ory Kratos + Hydra
- Separate Postgres instance for Ory with required extensions
- Tested end-to-end on Azure


azure/examples/enterprise/variables.tf
```diff
@@ -0,0 +1,100 @@
...
+  type        = string
+}
+
+variable "ory_oel_image_tag" {
+  description = "Image tag for OEL images."
+  type        = string
+  default     = "26.2.3"
+}
+
+# TODO: Update auth mechanism once Materialize private registry is set up.
```
bobbyiliev: The auth mechanism will need updating once the Materialize private registry/mirror is set up, marked with TODOs in the code.

azure/examples/enterprise/variables.tf
```diff
@@ -0,0 +1,136 @@
...
+variable "internal_load_balancer" {
+  description = "Whether to use an internal load balancer"
+  type        = bool
+  default     = true
+}
+
+variable "enable_observability" {
+  description = "Enable Prometheus and Grafana monitoring stack for Materialize"
+  type        = bool
+  default     = false
```
jubrad: Outside of this PR we may want to default this to true

azure/examples/enterprise/README.md
```diff
@@ -0,0 +1,285 @@
+# Example: Enterprise Materialize Deployment on Azure with Ory (OIDC/SAML)
+
+This example extends the [simple deployment](../simple/) with **Ory Kratos** (identity management) and **Ory Hydra** (OAuth2/OIDC provider) for enterprise authentication via OIDC and SAML.
+
+---
```
jubrad, line 7: Might be worth putting a note on here that this is WIP, until we get the registry proxy, and we get closer to releasing the feature

azure/examples/enterprise/main.tf
```diff
@@ -0,0 +1,976 @@
...
+    min_nodes            = 2
+    max_nodes            = 5
+    node_count           = null
+    disk_size_gb         = 100
+    swap_enabled         = true
+  }
+
+  database_config = {
+    sku_name                      = "GP_Standard_D2s_v3"
+    postgres_version              = "15"
```
jubrad: Why 15? 18 has been GA since september of last year?

azure/examples/enterprise/main.tf
```diff
@@ -0,0 +1,976 @@
...
+    storage_mb                    = 32768
+    backup_retention_days         = 7
```
jubrad: If we're making other "enterprise" decisions we should probably bump this to 60 days

azure/examples/enterprise/main.tf
```diff
@@ -0,0 +1,976 @@
...
+  # generates a UUID client_id; the installed CRD version doesn't support setting
+  # it explicitly).
+  # See: https://materialize.com/docs/security/self-managed/sso/
+  system_parameters = {
+    oidc_issuer               = local.hydra_external_url
+    oidc_audience             = jsonencode([data.kubernetes_secret_v1.oauth2_client.data["CLIENT_ID"]])
+    oidc_authentication_claim = "email"
+    console_oidc_client_id    = data.kubernetes_secret_v1.oauth2_client.data["CLIENT_ID"]
+    console_oidc_scopes       = "openid email"
+  }
```
jubrad: nice

It may also be worth noting this can also be used to set other configuration parameters
https://materialize.com/docs/sql/alter-system-set/#key-configuration-parameters

azure/examples/enterprise/main.tf
```diff
@@ -0,0 +1,976 @@
...
+    kubectl_manifest.ory_certificate["kratos-tls"],
+  ]
+}
+
+# TLS certificates for Hydra, Kratos and the selfservice UI. The optional
+# *.cluster.local SAN is dropped when the customer brings their own
+# (potentially public ACME) issuer that can't sign single-label cluster
+# names; in that case in-cluster callers route via the public hostname
+# (hairpin NAT through the LB; TLS still validates).
+resource "kubectl_manifest" "ory_certificate" {
```
jubrad: It seems to me like the example files should be largely be sticking together the the modules we've made with some glue. This is is starting to feel like the example is becoming mostly glue.

It might be worth looking for some opportunities to push items into modules.

A quick ask of claude to identify non module similarities between the azure, gcp, and aws enterprise examples:

  

> Azure: 976 lines, GCP: 946 lines, AWS: 1153 lines — and these raw resources are identical in all three:
> 
>   kubernetes_namespace.ory
>   kubernetes_secret.ory_oel_registry
>   kubectl_manifest.ory_certificate
>   kubernetes_service_v1.ory_lb
>   kubernetes_service_v1.console_lb_443
>   kubernetes_network_policy_v1.materialize_to_ory_egress
>   kubernetes_network_policy_v1.ory_from_materialize_ingress
>   kubectl_manifest.materialize_oauth2_client
>   data.kubernetes_secret_v1.oauth2_client
> 
>   That's ~300-400 lines of near-identical code living in three places that will drift independently. The right move is to pull the Ory-specific glue into a module. The natural seam is something like an ory-stack composite module that wraps Kratos +
>   Hydra + selfservice-ui together with their certs, LBs, namespace, registry secret, and OAuth2 client registration — the things that are always deployed together anyway and that no one would want to wire individually.
> 
>   The network policies are the one genuine exception — they cross the materialize-environment ↔ ory namespace boundary and need to live in the example or a separate "wiring" module that takes both namespace names as inputs.
> 
>   The console_lb_443 service is also worth looking at — it's not in the existing load_balancers module and is appearing separately in all three examples, which suggests the load_balancers module has an incomplete interface.
> 
>   This seems like the right time to raise it, before those other two PRs merge and the pattern gets set in stone.

jubrad (commented):
Overall I think the code looks pretty good, but I do worry about some srawl and the potentiall for these example modules to start really deviating from each other and the simple examples with how much is getting pushed into them. Might be woth a chat with @alex-hunt-materialize on if there's a better way to structure this.

azure/examples/enterprise/main.tf
```diff
@@ -0,0 +1,976 @@
...
+    name      = "ory-oel-registry"
+    namespace = kubernetes_namespace.ory.metadata[0].name
+  }
+
+  type = "kubernetes.io/dockerconfigjson"
+
+  data = {
+    ".dockerconfigjson" = jsonencode({
+      auths = {
+        "europe-docker.pkg.dev" = {
```
alex-hunt-materialize: This will need to change when we get the oci-registry-proxy deployed.

alex-hunt-materialize (commented):
Overall, this looks pretty good. I really like the detailed README.

I generally agree with Justin's comments. It would be really nice if we could have a single Ory module that does as much as possible of this. We may actually want multiple layers for that:

1. The individual kratos, hydra, etc... modules that aren't cloud-provider specific, in the kubernetes module dir.
2. A module that wraps all of the above for easy reuse in the different cloud providers.
3. A cloud-provider specific module that includes things that can't be added to the cloud agnostic module above, such as the DB instantiation.

I'm not 100% sure if it is worth having layer (2), but that's something that will become obvious during implementation.

This would allow this example to be much simpler, since it just needs to instantiate module (3).

azure/examples/enterprise/main.tf
bobbyiliev: Bumped to 35, which is the cap on built-in automated backups for Azure Flexible Server and AWS RDS.

azure/examples/enterprise/terraform.tfvars.example
```diff
@@ -0,0 +1,49 @@
...
+license_key         = "your-materialize-license-key"
+
+# Restrict access to the AKS API server. Required (no default). For lab use
+# you may pass ["0.0.0.0/0"], but production should pin a tight allowlist.
+k8s_apiserver_authorized_networks = ["203.0.113.0/24"]
+
+# OEL image pull credentials. Contact Materialize sales for access to Ory's
+# private registry while the Materialize-hosted mirror is in flight.
+ory_oel_registry = "europe-docker.pkg.dev/ory-artifacts"
+ory_oel_key_file = "/path/to/ory-artifacts-sa-key.json"
```
alex-hunt-materialize: These are left over after the change to use the proxy.
```suggestion
```

azure/examples/enterprise/main.tf
```diff
@@ -0,0 +1,568 @@
...
+  vnet_config = {
+    address_space                      = "20.0.0.0/16"
+    aks_subnet_cidr                    = "20.0.0.0/20"
+    postgres_subnet_cidr               = "20.0.16.0/24"
+    enable_api_server_vnet_integration = true
+    api_server_subnet_cidr             = "20.0.32.0/27" # keeping atleast 32 IPs reserved for API server and related services used in delegation might reduce it later.
+  }
+
+  aks_config = {
+    kubernetes_version         = "1.33"
```
alex-hunt-materialize:
```suggestion
    kubernetes_version         = "1.34"
```
We should match the simple example here. We probably want to bump both to 1.35 soon, but that can be out of scope of this example.

azure/examples/enterprise/main.tf
```diff
@@ -0,0 +1,568 @@
...
+    min_nodes            = 2
+    max_nodes            = 5
+    node_count           = null
+    disk_size_gb         = 100
+    swap_enabled         = true
+  }
+
+  database_config = {
+    sku_name                      = "GP_Standard_D2s_v3"
+    postgres_version              = "17"
```
alex-hunt-materialize, line 72: Why 17 instead of 18? This also applies below.

azure/examples/enterprise/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+```bash
+az keyvault purge --name <vault-name> --location <location>
+```
+
+For lab iteration where you do not care about recovery, you can flip `purge_soft_delete_on_destroy = true` in the `azurerm` provider block.
+
+### Balancerd SAN
+
+This example only puts `materialize_console_hostname` into the **console** cert SAN list. Balancerd (the SQL wire-protocol endpoint) sits behind its own LoadBalancer and is not exposed under a public hostname by default. If you want external SQL access, register a separate hostname, create an A record for it pointing at the balancerd LB IP, and add it to the `balancerd_extra_dns_names` argument of the `materialize_instance` module call in `main.tf`.
```
alex-hunt-materialize: I don't think an externally accessed console will be functional without the balancerd hostname, cert, and A record. When accessing the console through the browser, it needs to connect to balancerd. If the console only has an untrusted internal hostname for that, it won't work.

alex-hunt-materialize (commented):
Overall, this looks pretty good.

azure/examples/enterprise/main.tf
bobbyiliev: 18 isn't GA on Azure Database for PostgreSQL Flexible Server yet, the apply errored when I tried it during e2e testing so I dropped both `database_config` and `ory_database_config` back to 17. Happy to bump once Azure makes 18 GA.

azure/examples/enterprise/terraform.tfvars.example
bobbyiliev: Good catch, dropped both lines.

azure/examples/enterprise/main.tf
bobbyiliev: Bumped to 1.34 to match the simple example. Agreed on 1.35 as a follow-up across both!

azure/examples/enterprise/main.tf
bobbyiliev: 18 isn't GA on Azure Database for PostgreSQL Flexible Server yet, the apply errored when I tried it during e2e testing so I dropped both `database_config` and `ory_database_config` back to 17. Happy to bump once Azure makes 18 GA.

azure/examples/enterprise/README.md
bobbyiliev: You're right, fixed!

azure/examples/enterprise/main.tf
alex-hunt-materialize: It is GA? https://techcommunity.microsoft.com/blog/adforpostgresql/postgresql-18-now-ga-on-azure-postgres-flexible-server/4469802

azure/examples/enterprise/main.tf
alex-hunt-materialize: It is GA, though. https://techcommunity.microsoft.com/blog/adforpostgresql/postgresql-18-now-ga-on-azure-postgres-flexible-server/4469802

azure/examples/enterprise/README.md
```diff
@@ -67,7 +68,7 @@ tags = {
 - `tags`: Map of tags to apply to resources
 - `license_key`: Materialize license key JWT. Used for Materialize itself and as the password authenticating to the Ory registry proxy. Must carry the `ory` entitlement.
 - `k8s_apiserver_authorized_networks`: List of CIDR blocks allowed to reach the AKS API server. No default; pass `["0.0.0.0/0"]` for lab use, or a tight allowlist for production.
-- `ory_hydra_hostname`, `ory_ui_hostname`, `ory_kratos_hostname`, `materialize_console_hostname`: Public hostnames for the four browser-facing services
+- `ory_hydra_hostname`, `ory_ui_hostname`, `ory_kratos_hostname`, `materialize_console_hostname`, `materialize_balancerd_hostname`: Public hostnames for the five browser-facing endpoints (Hydra OAuth2, Kratos public API, selfservice UI, Materialize console, and balancerd — the SQL wire endpoint the console JS calls from the browser)
```
alex-hunt-materialize, line 71: Nitpick, it's SQL over HTTP.

alex-hunt-materialize (approved):
Postgresql 18 is GA on Azure, but I won't block this over that.
https://techcommunity.microsoft.com/blog/adforpostgresql/postgresql-18-now-ga-on-azure-postgres-flexible-server/4469802

PR #168, opened 2026-03-15
Add GCP enterprise example with Ory Kratos and Hydra for OIDC/SAML authentication DEP-141

bobbyiliev:
Same as https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/167 but for GCP


SangJunBak (dismissed):
Everything looks quite logical, though I'm out of my depth for terraform reviews 😆 . One question I had was I know Ory Kratos Open source is separate from the Ory enterprise license. Do we need to put anything like a license key into any of the ory modules?

Screenshot 2026-03-17 at 5 54 20 PM

Not sure if this should go in this PR but realistically, most users are going to want to plug an upstream IdP (e.g. Okta) into Kratos. Wondering if there should be some sort of example for this.

alex-hunt-materialize (approved):

PR #169, opened 2026-03-15
Add AWS enterprise example with Ory Kratos and Hydra for OIDC/SAML authentication DEP-142

bobbyiliev:
Same as https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/167 but for AWS


aws/examples/enterprise/.terraform.lock.hcl
```diff
@@ -1,203 +0,0 @@
-# This file is maintained automatically by "terraform init".
```
alex-hunt-materialize, line 1: Nitpick, if we don't want this file committed, we should add it to .gitignore.

aws/examples/enterprise/main.tf
```diff
@@ -0,0 +1,732 @@
...
+  vpc_id                    = module.networking.vpc_id
+  cluster_name              = module.eks.cluster_name
+  cluster_security_group_id = module.eks.cluster_security_group_id
+  node_security_group_id    = module.eks.node_security_group_id
+
+  tags = var.tags
+}
+
+# Separate RDS instance for Ory Hydra
+module "ory_hydra_database" {
```
alex-hunt-materialize, line 386: The Azure PR had only one postgres instance for both kratos and hydra, but here we're creating two. Is there a reason for that?

You don't need to change anything here, I just want to understand the choices. Does our AWS database module just not support multiple databases per postgres instance?

alex-hunt-materialize (dismissed):

aws/examples/enterprise/.terraform.lock.hcl
bobbyiliev: Good catch, added `.terraform.lock.hcl` to the `.gitignore` so it doesn't drift back in.

aws/examples/enterprise/main.tf
bobbyiliev: Yeah, that's the limitation. On Azure there's a separate `azurerm_postgresql_flexible_server_database` resource, so the module just takes a list of databases and creates both on one instance.

The AWS provider has no equivalent, eg. the `aws_db_instance` only creates the initial `db_name`, anything beyond that means running `CREATE DATABASE` against the live RDS. So landed on two RDS instances. We've had that issue in the past when we were using 1 RDS instance for multiple Materialize installations as well and we had a Kubernetes Job run the `CREATE DATABASE` statements, so we could do something similar here too in a follow up PR.

alex-hunt-materialize: Yeah, if we're having to manage SQL commands directly, it's probably better to keep it with separate instances.

alex-hunt-materialize (approved):

PR #186, opened 2026-04-09
Add OEL image support to Ory Kratos and Hydra modules

bobbyiliev:
Adding those vars to ory-kratos and ory-hydra modules to support Ory Enterprise License deployments with private registry images.


alex-hunt-materialize (commented):
This all looks fine, but doesn't match my design doc https://github.com/MaterializeInc/cloud/pull/12390

bobbyiliev:
> This all looks fine, but doesn't match my design doc [MaterializeInc/cloud#12390](https://github.com/MaterializeInc/cloud/pull/12390)

Yep, this is intentionally ahead of the design doc, the module changes here just add image_repository, image_tag, and image_pull_secrets variables, which are needed regardless of the auth mechanism.

I've tested OEL images end-to-end on all three clouds using a temporary GCP key file approach to make sure it all works together.

Will adapt to match the design doc (JWT parsing, secret creation inside modules) in a follow-up once we have it approved! Should be a straightforward swap on top of this 🤞

alex-hunt-materialize (approved):

alex-hunt-materialize (dismissed):

alex-hunt-materialize (approved):

PR #187, opened 2026-04-09
Add Ory Polis module for SAML-to-OIDC bridge and SCIM directory sync

bobbyiliev:
Adding Ory Polis Terraform module for deploying Polis as a SAML-to-OIDC bridge
and SCIM directory sync service. Polis would be required for customers who need to
federate with SAML identity providers (Okta, Azure AD, etc.) as it basically translates
SAML assertions into standard OAuth 2.0 flows that Hydra and Kratos can consume.

Deployed using raw Kubernetes resources (Deployment + Service) since Ory does
not publish an official Polis Helm chart yet.


kubernetes/modules/ory-polis/variables.tf
```diff
@@ -0,0 +1,147 @@
...
+  description = "Docker image registry for Polis. Example: europe-docker.pkg.dev"
+  type        = string
+  default     = "docker.io"
+  nullable    = false
+}
+
+variable "image_repository" {
+  description = "Docker image repository for Polis. Example: ory-artifacts/ory-enterprise-polis/polis-oel"
+  type        = string
+  default     = "boxyhq/jackson"
```
alex-hunt-materialize: Is this correct? What is this repository?

kubernetes/modules/ory-polis/outputs.tf
```diff
@@ -0,0 +1,26 @@
+output "url" {
+  description = "Internal cluster URL for Polis (SSO, SCIM, OAuth endpoints). Always http because Polis does not serve TLS itself — TLS is terminated externally in front of the pod."
```
alex-hunt-materialize: What is going to be doing the TLS termination?

kubernetes/modules/ory-polis/variables.tf
bobbyiliev: Good catch, this needed a better docstring. Ory Polis is actually a rebrand/fork of BoxyHQ's SAML Jackson project, Ory took stewardship and the source now lives at https://github.com/ory/polis, but the OSS Docker image is still published to the upstream `boxyhq/jackson` repo on Docker Hub (there is no `oryd/polis` image). Release tags track the Polis cadence, e.g. `v26.2.0` maps to `boxyhq/jackson:26.2.0`.

I've expanded the variable description to explain this, and bumped the default tag from the old BoxyHQ-era `1.52.2` to the current `26.2.0`. OEL users would override both this variable and `image_registry` with the Ory private registry path (e.g., `europe-docker.pkg.dev/ory-artifacts/ory-enterprise-polis/polis-oel`).

kubernetes/modules/ory-polis/outputs.tf
bobbyiliev: Polis itself does not terminate TLS, it's a NextJS app that only speaks plain HTTP on its configured port (5225 by default). TLS termination is the caller's responsibility, typically done by whatever fronts the pod in the enterprise example wiring we've got a cloud LoadBalancer service with a cert-manager cert (the pattern we use for Hydra/Kratos/selfservice-ui)

I've expanded the `url` output description to make that contract explicit, and the `external_url` variable description already calls out that it must be the HTTPS URL of whatever proxy terminates TLS in front of Polis.

kubernetes/modules/ory-polis/outputs.tf
```diff
@@ -0,0 +1,26 @@
+output "url" {
```
alex-hunt-materialize: Nitpick, maybe `internal_url` would be better, to match with the `external_url` we also have.

alex-hunt-materialize (dismissed):

kubernetes/modules/ory-polis/main.tf
```diff
@@ -0,0 +1,160 @@
...
+          cpu    = var.resources.requests.cpu
+          memory = var.resources.requests.memory
+        }
+        limits = merge(
+          { memory = var.resources.limits.memory },
+          var.resources.limits.cpu != null ? { cpu = var.resources.limits.cpu } : {}
+        )
+      }
+    }
+  }, local.image_config, local.image_pull_secrets_config, local.monitoring_config)
```
alex-hunt-materialize, line 139: Does this automatically roll the deployment on changes to the secret? I'm not sure it does, and claude thinks it doesn't.

Perhaps we should add an annotation to the deployment's pod template? Maybe something like
```
podAnnotations.checksum = sha256(jsonencode(kubernetes_secret.polis.data))
```

I think this might apply to all the Ory modules. This isn't catastrophic, but we probably want to fix it.

kubernetes/modules/ory-polis/outputs.tf
```diff
@@ -0,0 +1,42 @@
...
```
alex-hunt-materialize: I still feel that `internal_url` would be a better name.

alex-hunt-materialize (dismissed):
A couple of small issues, but nothing I'd block this for.

kubernetes/modules/ory-polis/outputs.tf
bobbyiliev: Good call, renamed to `internal_url`!

kubernetes/modules/ory-polis/outputs.tf
bobbyiliev: Yep, done in the last commit, thanks for the nudge!

kubernetes/modules/ory-polis/main.tf
bobbyiliev: Good catch, added a `checksum/config` pod annotation derived from the secret hash, and you're right that the other Ory modules have the same gap so I'll open a follow-up PR to fix them in the same shape.

alex-hunt-materialize (approved):

PR #188, opened 2026-04-09
Regenerate module README docs via terraform-docs

bobbyiliev:
Just ran the `./.github/scripts/generate-docs.sh` and noticed some outdated README files.


alex-hunt-materialize (approved):

alex-hunt-materialize:
We should probably add a lint for this.

PR #200, opened 2026-04-27
Add Let's Encrypt ClusterIssuer module (Cloudflare DNS-01)

bobbyiliev:
Adding the `letsencrypt-cluster-issuer` module so the per-cloud Ory enterprise examples can opt into a real ACME issuer (Cloudflare DNS-01 to start, more providers can plug into the `dns_provider` enum later).

### DIFF
--- a/kubernetes/modules/letsencrypt-cluster-issuer/README.md
+++ b/kubernetes/modules/letsencrypt-cluster-issuer/README.md
@@ -10,8 +10,8 @@
 
 | Name | Version |
 |------|---------|
-| <a name="provider_kubectl"></a> [kubectl](#provider\_kubectl) | 2.2.0 |
-| <a name="provider_kubernetes"></a> [kubernetes](#provider\_kubernetes) | 3.1.0 |
+| <a name="provider_kubectl"></a> [kubectl](#provider\_kubectl) | ~> 2.0 |
+| <a name="provider_kubernetes"></a> [kubernetes](#provider\_kubernetes) | >= 2.10.0 |
 
 ## Modules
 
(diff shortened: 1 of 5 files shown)


alex-hunt-materialize (commented):
Is this really easier than just letting people define whatever they want in their own issuers in `kubectl_manifest`? This is a bunch of extra code which is much less flexible than just directly instantiating the kubectl_manifest.

bobbyiliev:
> Is this really easier than just letting people define whatever they want in their own issuers in `kubectl_manifest`? This is a bunch of extra code which is much less flexible than just directly instantiating the kubectl_manifest.

Yea true, what I had in mind initially was that the module bundles the Cloudflare API token Secret with the ClusterIssuer manifest, validates the inputs (`acme_environment`, `dns_provider`), and surfaces a typed interface so a customer sees `letsencrypt_email`, `cloudflare_api_token` etc. instead of writing the YAML themselves.

But you're right that it's locked to Cloudflare DNS-01 today, and the enterprise example already has a `cert_issuer_ref` BYO path, so the niche it covers is narrow. The alternative would be to drop the module and put a copyable LE-with-Cloudflare `kubectl_manifest` snippet in the enterprise README instead. So indeed more flexible, less code to maintain.

What do you think?

alex-hunt-materialize:
Yeah, I'd just give an example and let them figure it out for their situation. We really have no idea which DNS-01 issuer they will want to use.

bobbyiliev:
Closing per the discussion above. Going to drop the module and document a copyable Let's Encrypt with Cloudflare DNS-01 snippet in the enterprise example README instead. The example continues to support BYO via `cert_issuer_ref`, so customers can paste the snippet (or any DNS-01 / HTTP-01 solver they prefer) into their root module and pass the issuer name through.

Updating PRs #167 / #168 / #169 to drop the module wiring.

PR #204, opened 2026-05-18
Add ory-stack composite module

bobbyiliev:
As discussed in https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/167


kubernetes/modules/ory-stack/main.tf
```diff
@@ -0,0 +1,497 @@
+locals {
+  # When materialize_namespace is set, the module wires Materialize-side resources
+  # (OAuth2 client, network policies, console LB). When null, those are skipped.
+  wire_materialize = var.materialize_namespace != null
+
+  # External URLs that the browser (and Materialize, for OIDC issuer matching)
+  # sees. Hostnames resolve to the LB IPs and are terminated by cert-manager certs.
+  hydra_external_url  = "https://${var.hydra_hostname}/"
+  kratos_external_url = "https://${var.kratos_hostname}"
+  ui_external_url     = "https://${var.ui_hostname}"
```
alex-hunt-materialize: Should these be FQDNs?

kubernetes/modules/ory-stack/main.tf
```diff
@@ -0,0 +1,497 @@
...
+
+  # Cookie domain shared across the Ory subdomains so flow/session cookies work
+  # across sibling hostnames (Kratos, UI, Hydra). Defaults to the parent domain
+  # of kratos_hostname (e.g. kratos.example.com -> example.com).
+  cookie_parent_domain = var.cookie_parent_domain != null ? var.cookie_parent_domain : regex("^[^.]+\\.(.+)$", var.kratos_hostname)[0]
```
alex-hunt-materialize: Am I reading this right? Is this regex meant to just split on the first `.`? If so, terraform has `split` and `join` functions that might be more readable.

Also, if the user passes only a hostname (as the variable name implies), this will error on not finding the `.`. We should probably rename these variables AND not use the regex here.

kubernetes/modules/ory-stack/main.tf
```diff
@@ -0,0 +1,497 @@
+locals {
+  # When materialize_namespace is set, the module wires Materialize-side resources
+  # (OAuth2 client, network policies, console LB). When null, those are skipped.
+  wire_materialize = var.materialize_namespace != null
+
+  # External URLs that the browser (and Materialize, for OIDC issuer matching)
+  # sees. Hostnames resolve to the LB IPs and are terminated by cert-manager certs.
+  hydra_external_url  = "https://${var.hydra_hostname}/"
```
alex-hunt-materialize: This is the only one with a trailing `/`. If these are used for issuer matching, won't that be an exact string match? Shouldn't all of these follow the same convention?

kubernetes/modules/ory-stack/main.tf
```diff
@@ -0,0 +1,497 @@
...
+# OAuth2 client via Hydra's admin API. The Secret named here is populated by
+# Hydra Maester with the generated client_id and client_secret.
+resource "kubectl_manifest" "materialize_oauth2_client" {
+  count = local.wire_materialize ? 1 : 0
+
+  yaml_body = yamlencode({
+    apiVersion = "hydra.ory.sh/v1alpha1"
+    kind       = "OAuth2Client"
+    metadata = {
+      name      = "materialize"
```
alex-hunt-materialize: Hard-coded name. Shouldn't it be prefixed or something?

What if they have multiple materialize instances in the same namespace? Will this have them all share the same Ory installation?

kubernetes/modules/ory-stack/main.tf
```diff
@@ -0,0 +1,497 @@
...
+      "app.kubernetes.io/instance" = each.value.app_instance
+    }
+
+    port {
+      name        = "https"
+      port        = 443
+      target_port = each.value.target_port
+      protocol    = "TCP"
+    }
+  }
```
alex-hunt-materialize, line 336: Should this also have `wait_for_load_balancer = true`, like the console one?

kubernetes/modules/ory-stack/main.tf
```diff
@@ -0,0 +1,497 @@
...
+        }
+      }
+    }
+
+    # External traffic from the LBs hits Hydra public (4444), Kratos public
+    # (4433), and the selfservice UI (3000). Admin ports stay internal.
+    ingress {
+      from {
+        ip_block {
+          cidr = "0.0.0.0/0"
```
alex-hunt-materialize: Should this be a passed in variable?

kubernetes/modules/ory-stack/main.tf
```diff
@@ -0,0 +1,497 @@
...
+  }
+}
+
+# Allow Ory pods to receive traffic from Materialize, from within the ory
+# namespace, and from external LoadBalancers on the three public ports.
+resource "kubernetes_network_policy_v1" "ory_from_materialize_ingress" {
+  count = local.wire_materialize ? 1 : 0
+
+  metadata {
+    name      = "allow-materialize-ingress"
```
alex-hunt-materialize: This isn't just from materialize.

kubernetes/modules/ory-stack/variables.tf
```diff
@@ -0,0 +1,224 @@
...
+variable "create_namespace" {
+  description = "Whether the module should create the Ory namespace. Set to false if it already exists."
+  type        = bool
+  default     = true
+  nullable    = false
+}
+
+# Hostnames -------------------------------------------------------------------
+
+variable "hydra_hostname" {
```
alex-hunt-materialize: We should rename these and update their descriptions to make it clear that they should be FQDNs, not hostnames.

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Yep, indeed they should be. I just renamed them.

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Good call. Last commit replaces the regex with a `split` / `join` over `kratos_fqdn`, and falls back to the fqdn itself when there's no `.` instead of erroring.

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Agreed. I just dropped the trailing slash on `hydra_external_url` in the last commit.

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Hm yes, fixed in the last commit.

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Added in the last commit!

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Good point, made it a variable in the last commit.

kubernetes/modules/ory-stack/variables.tf
bobbyiliev: Done in the last commit.

alex-hunt-materialize (approved):

PR #210, opened 2026-05-22
Pin all dependency versions and configure dependabot

def-:
Add upper-bound caps to every Terraform provider constraint, pin GitHub Actions to commit SHAs (with `# vX.Y.Z` comments), and tighten loose Cargo constraints to exact lock versions. Add `.github/dependabot.yml` that groups all updates per ecosystem (terraform, github-actions, cargo) into a single PR with a 7-day cooldown.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,84 @@
+version: 2
+updates:
+  - package-ecosystem: "terraform"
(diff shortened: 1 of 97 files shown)


bobbyiliev (approved):
Thanks @def-!

PR #222, opened 2026-06-02
Default enable_observability to true in the simple examples

bobbyiliev:
Following up on https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/167#discussion_r3231578459.


alex-hunt-materialize (approved):
@Alphadelta14 We may need to do migrations once your monitoring stack changes are ready. These could already have been enabled manually, so we have to do that regardless of this change to the default.

jubrad (approved):

PR #223, opened 2026-06-02
Add optional Polis support to ory-stack and wire it into the Azure enterprise example DEP-145

bobbyiliev:
Adding an `enable_polis` flag to `ory-stack` for customers on SAML-only IdPs and for uniform SCIM provisioning. Defaulted off.

Also wires the Azure enterprise example so you can flip it on via tfvars.

Two Polis-side rough edges worth flagging:
- `polis-oel` image is amd64-only (Kratos/Hydra are multi-arch). On arm64 clusters we need a small dedicated amd64 nodepool for Polis. Pinged Ory for a multi-arch manifest.
- polis chart is private OEL and our OEL registry proxy doesn't serve OCI chart manifests yet (image pulls work fine, chart manifests 403). Workaround is a GCP SA key for the chart pull (`ory_polis_oci_chart_key_file`). Pinged Ory about publishing publicly and the proxy team about adding chart support, either fix drops the SA key path.

Other smaller bits the module handles inside: nginx TLS proxy in front (Polis doesn't terminate TLS), auto-generated RSA keys for OIDC token signing, `EXTERNAL_URL` env var so SCIM endpoints aren't `localhost`. Also flagged this to Ory.

Tested end to end on Azure: SCIM push from Okta into Polis works, full SSO chain (console -> Hydra -> Kratos -> Polis -> Okta -> back) lands logged in.

I'll file separately the same wiring for GCP and AWS examples.

Open questions around mapping SCIM group claims to SQL grants? Could fall under the [SCIM role mapping project](https://linear.app/materializeinc/project/scim-role-mapping-b9d87ea047c4/issues)? Polis side is ready to surface groups in the OIDC token whenever the consumer side picks it up.


azure/examples/enterprise/variables.tf
```diff
@@ -98,6 +98,31 @@ variable "ory_kratos_hostname" {
...
 }
 
+variable "enable_polis" {
+  description = "Deploy Ory Polis (SAML-to-OIDC bridge) alongside Kratos and Hydra. When true, ory_polis_hostname must be set and a polis database is provisioned on the Ory Postgres instance."
+  type        = bool
+  default     = false
+  nullable    = false
+}
+
+variable "ory_polis_hostname" {
```
alex-hunt-materialize: I thought we named all these with `_fqdn` endings, but maybe this slipped through. If it's an FQDN, we should name the vars FQDN.

You don't have to fix this one now, as it seems all of them are named this way :(

azure/examples/enterprise/versions.tf
```diff
@@ -26,5 +26,13 @@ terraform {
...
       version = "2.4.0"
     }
+    # Declared so users can drop in a local okta.tf (gitignored) to automate
+    # the Okta SAML app + assignments for the Polis SCIM/SAML e2e test. Not
+    # used unless that file is present, in which case provider "okta" is
+    # configured there.
+    okta = {
+      source  = "okta/okta"
+      version = "~> 4.0"
+    }
```
alex-hunt-materialize, line 36: We should add support for this in the rust tests. Out of scope for this PR.

kubernetes/modules/ory-stack/main.tf
```diff
@@ -313,7 +338,206 @@ module "ory_selfservice_ui" {
...
+          "checksum/config" = sha256(kubernetes_config_map_v1.polis_tls_proxy[0].data["nginx.conf"])
+        }
+      }
+
+      spec {
+        node_selector = var.node_selector
+
+        container {
+          name              = "nginx"
+          image             = "nginxinc/nginx-unprivileged:1.27-alpine"
```
alex-hunt-materialize: We've seen lots of problems with DNS when using nginx in Kubernetes. This will probably work until the polis container gets recreated, at which point it will likely fail.

We fought with Nginx for years in our loki-gateways before we finally gave up and replaced it with https://github.com/vicanso/pingap

alex-hunt-materialize (dismissed):
We probably shouldn't use Nginx, but I won't block the PR for that.

azure/examples/enterprise/variables.tf
bobbyiliev: I decided to do this now before we've officially released all of it. So renamed across all three enterprise examples. The variables are now consistent with the FQDN naming the modules already use.

azure/examples/enterprise/versions.tf
bobbyiliev: Good point. Filed as [DEP-148](https://linear.app/materializeinc/issue/DEP-148/add-okta-provider-support-to-the-rust-tests-for-the-terraform-examples).

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Yes, makes sense! Just swapped to Pingap in the last commit, kept the same Deployment / ConfigMap / Service shape so the LB selector and port stay the same.

kubernetes/modules/ory-polis/README.md
```diff
@@ -57,7 +59,7 @@ No modules.
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_admin_api_keys"></a> [admin\_api\_keys](#input\_admin\_api\_keys) | Bearer token for authenticating requests to Polis admin APIs (set as the API\_KEYS env var inside the Polis container). If null, a random 32-character key is generated. | `string` | `null` | no |
-| <a name="input_chart_registry"></a> [chart\_registry](#input\_chart\_registry) | OCI registry hostname for the Polis Helm chart. | `string` | `"europe-west3-docker.pkg.dev"` | no |
+| <a name="input_chart_registry"></a> [chart\_registry](#input\_chart\_registry) | OCI registry hostname for the Polis Helm chart. Defaults to the EU multi-region Artifact Registry endpoint, which Ory mirrored the chart onto. The single-region 'europe-west3-docker.pkg.dev' path still works if you need it. | `string` | `"europe-docker.pkg.dev"` | no |
```
alex-hunt-materialize: Nitpick, this looks like a claude-injected explanation for the author, rather than something user-facing.

kubernetes/modules/ory-polis/variables.tf
```diff
@@ -28,9 +28,9 @@ variable "dsn" {
 
 # Helm chart source
 variable "chart_registry" {
-  description = "OCI registry hostname for the Polis Helm chart."
+  description = "OCI registry hostname for the Polis Helm chart. Defaults to the EU multi-region Artifact Registry endpoint, which Ory mirrored the chart onto. The single-region 'europe-west3-docker.pkg.dev' path still works if you need it."
```
alex-hunt-materialize: Same here.

alex-hunt-materialize (dismissed):

alex-hunt-materialize (approved):

PR #225, opened 2026-06-03
switch to Standard_D4ps_v5 for azure default node pool

alex-hunt-materialize:
Switch to Standard_D4ps_v5 for azure default node pool.

It is likely more widely available, we don't need local NVME disks for the default node pool, and there is nothing performance critical on those nodes.

This should fix the availability issues seen in https://github.com/MaterializeInc/materialize-terraform-self-managed/actions/runs/26824693413
Resolves https://linear.app/materializeinc/issue/CLO-88/azure-ci-aks-node-vm-size-capacity-failure


bobbyiliev (approved):
Thanks Alex!

PR #233, opened 2026-06-04
Upgrade to grafana 13, use community helm charts

Alphadelta14:
This upgrades the Grafana helm chart to use the grafana-community charts (which is their new home as of Jan 2026). I support changing it back via variable if needed, but generally this should be a good default.
This also upgrades the chart to 12.4.1 (which deploys Grafana 13.0.1-sec01).
This has been verified on a TF self-managed AWS/EKS cluster.

### DIFF
--- a/kubernetes/modules/grafana/README.md
+++ b/kubernetes/modules/grafana/README.md
@@ -1,7 +1,7 @@
 ## Requirements
 
 | Name | Version |
-|------|---------|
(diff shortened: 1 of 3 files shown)


bobbyiliev (dismissed):
Looks good! There are just some failing lints because of the comment in the `chart_repository` variable.

bobbyiliev (approved):

PR #241, opened 2026-06-15
Set default dashboard to newer materialize-monitoring dashboard

Alphadelta14:
This sets the default dashboard in Grafana to the newer env-top Dashboard from the materialize-monitoring project.

This is manually synced for now (we will eventually replace this with a helm module that contains this).

The dashboards can be found here: https://materializeinc.github.io/materialize-monitoring/dashboards/grafana/


alex-hunt-materialize (approved):
I cannot reasonably review the contents of a generated 7000+ line json dashboard, so I did not attempt to.

PR #246, opened 2026-06-26
Add materialize-terraform-self-managed agent skill

bobbyiliev:
Agent skill covering the Terraform modules for deploying self-managed Materialize on AWS, Azure, and GCP. Linear: DEX-36.

### DIFF
--- a/.claude/skills/materialize-terraform-self-managed/SKILL.md
+++ b/.claude/skills/materialize-terraform-self-managed/SKILL.md
@@ -262,26 +262,72 @@ region      = "us-central1"           # optional, defaults to us-central1
 | `crd_version` | `v1alpha1` | Use `v1` for v26.30+ |
 | `enable_observability` | `false` | Install Prometheus + Grafana |
 
-## Connecting After Deployment
+## Post-Deployment Setup
+
+### Configure kubectl
+
+After `terraform apply`, configure kubectl to talk to the new cluster:
+
+**AWS:**
+```bash
+aws eks update-kubeconfig --name $(terraform output -raw eks_cluster_name) --region <region>
+```
+
+**Azure:**
+```bash
+az aks get-credentials --resource-group <rg> --name $(terraform output -raw aks_cluster_name)
+```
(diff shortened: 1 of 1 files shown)


.claude/skills/materialize-terraform-self-managed/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+Set via `crd_version` variable in the example or the materialize-instance module.
+
+## Rollout Strategies
+
+| Strategy | Behavior | Resource impact |
+|----------|----------|-----------------|
+| `WaitUntilReady` | New pods created, cutover when healthy | Temporarily doubles resources |
+| `ManuallyPromote` | New pods created, waits for manual `forcePromote` | Temporarily doubles resources |
+| `ImmediatelyPromoteCausingDowntime` | Old pods torn down first | No extra resources, causes downtime |
```
jubrad, line 210: We can include some info here, but we may also want to link to our docs for more details
https://materialize.com/docs/self-managed-deployments/upgrading/

If there's a way to link to the raw md files that might be less token intensive.

.claude/skills/materialize-terraform-self-managed/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+Azure Blob:
+```
+https://storageaccount.blob.core.windows.net/container
+```
+
+GCP Cloud Storage (via HMAC/S3-compatible API):
+```
+s3://hmac-access-id:hmac-secret@bucket-name/materialize?endpoint=https%3A%2F%2Fstorage.googleapis.com&region=us-central1
+```
```
jubrad, line 194: Worth noting more explicitly that currently MZ only supports IAM auth with AWS S3 and Azure.

jubrad (commented):
It'd be nice to have a plan to keep this up-to-date. 


I also think we should include some directions for how to pull in these modules and update them in an existing terraform project, we can look at what context graph is doing there.

PR #249, opened 2026-06-29
Install v1 Materialize CRD

alex-hunt-materialize:
Install v1 Materialize CRD by default.


kay-kim (dismissed):

Alphadelta14 (dismissed):
lgtm assuming helm will pass args properly (unsure if this silently works if installV1CRD doesn't exist).

bobbyiliev (dismissed):

val-materialize (dismissed):

kay-kim (commented):
oh wait ... sorry .. Shouldn't we still default to false?
That is, 
1) We want to have the fix so that people can use v1 but still default to v1alpha
2) Some discussion between engineering and product to see when we want to make change the default to be v1

val-materialize (approved):

PR #250, opened 2026-06-29
purge subcommand for AWS tests

alex-hunt-materialize:
Adds a new purge subcommand for the tests, to clean up AWS objects. At present, this is only for AWS. We may port the GCP and Azure cleanup code to this purge command later.

Resolves https://linear.app/materializeinc/issue/CLO-85/aws-ci-guarantee-test-resource-cleanup-when-terraform-destroy-fails

Tested locally and in https://github.com/MaterializeInc/materialize-terraform-self-managed/actions/runs/28440008423/job/84279054281 by cancelling the job mid apply.

### DIFF
--- a/test/src/commands/purge.rs
+++ b/test/src/commands/purge.rs
@@ -1,0 +1,2244 @@
+//! Tag-scoped deletion of leaked AWS resources for a test run.
+//!
+//! When the AWS infrastructure test's `terraform destroy` fails, resources are
+//! leaked, costing money and risking name/quota collisions on future runs. AWS
+//! has no single-container delete (no resource-group/project equivalent), so
+//! cleanup is tag-driven.
+//!
+//! Every taggable resource the tests create is stamped with a unique per-run
+//! tag: the terraform provider sets `default_tags { tags = var.tags }` and the
+//! harness populates `var.tags` with `TestRun=<test_run_id>` (see
+//! [`crate::commands::init`]). The same `<test_run_id>` is the terraform
+//! `name_prefix`, so the EKS cluster is named `<test_run_id>-eks`. This module
+//! uses that id to find and delete everything belonging to a run, independent
+//! of terraform state.
+//!
+//! ## Deletion order: a dependency DAG
+//!
+//! Resources are deleted respecting their teardown dependencies, running
+//! independent subsystems concurrently and blocking only where a deletion must
+//! finish before downstream cleanup can proceed:
+//!
+//! ```text
+//!   node groups ──wait──▶ terminate instances ──┬─wait─▶ EKS cluster ─wait─▶ logs ┐
+//!                                               ├──────▶ EBS volumes              │
+//!                                               └──────▶ launch templates         │
+//!   RDS instance ──wait──▶ RDS subnet group / parameter group ────────────────────┤
+//!   NAT gateways ──wait──▶ release EIPs ──────────────────────────────────────────┤
+//!   load balancers ───────────────────────────────────────────────────────────────┤
+//!   VPC endpoints / S3 / SQS+EventBridge / KMS (fully independent) ───────────────┤
+//!                                                                                 ▼
+//!         ENIs ─▶ security groups ─▶ subnets ─▶ route tables ─▶ IGW ─▶ VPC   (+ IAM)
+//! ```
+//!
+//! The cluster log group is deleted at the tail of the compute branch rather
+//! than in a parallel branch: the EKS control plane recreates it while the
+//! cluster drains, so it must be deleted only after the cluster-deletion wait.
+//!
+//! The blocking waits matter because each gates the network teardown: a managed
+//! node group's ASG would relaunch terminated instances; the EKS cluster's
+//! control-plane ENIs and the RDS instance both sit in the subnets; NAT gateways
+//! hold the subnets and their EIPs. Built-in SDK deletion waiters are used for
+//! these. The whole sweep is repeated a few times to mop up resources (e.g. ENIs
+//! the load balancer controller releases lazily) that only become deletable once
+//! their dependents are gone; once the big blockers are deleted, later passes
+//! return from the waiters immediately.
+//!
+//! ## Error handling
+//!
+//! Each branch is best-effort and distinguishes two kinds of failure:
+//!
+//! * A **deletion** failure (one resource of many) must not stop the branch from
+//!   deleting the rest, so [`handle`]/[`handle_wait`] record it in a per-branch
+//!   error list and carry on. "Already gone" is treated as success.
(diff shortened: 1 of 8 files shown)


test/src/commands/purge.rs
```diff
@@ -0,0 +1,2244 @@
...
+                .name(&self.cluster)
+                .wait(CLUSTER_DELETE_WAIT)
+                .await,
+        );
+        Ok(errors)
+    }
+
+    /// Deletes available EBS volumes: terraform-tagged (`TestRun`) and EBS CSI
+    /// driver PVs (tagged `kubernetes.io/cluster/<cluster>=owned`). Depends on
+    /// instances being terminated so the volumes have detached.
```
Alphadelta14, line 426: is this defensive enough to handle if a new instance comes up and reattaches the volume?

test/src/commands/purge.rs
```diff
@@ -0,0 +1,2244 @@
...
+                if !customer_managed || pending {
+                    continue;
+                }
+                let Ok(tags) = self.kms.list_resource_tags().key_id(key_id).send().await else {
+                    continue;
+                };
+                if tags
+                    .tags()
+                    .iter()
+                    .any(|t| t.tag_key() == "TestRun" && t.tag_value() == self.run_id)
```
Alphadelta14, line 906: should the condition be a function?

test/src/commands/purge.rs
```diff
@@ -0,0 +1,2244 @@
...
+                .any(|t| t.key() == "TestRun" && t.value() == self.run_id);
+            if !is_match {
+                continue;
+            }
+            self.empty_and_delete_bucket(&name, &mut errors).await;
+        }
+        Ok(errors)
+    }
+
+    async fn empty_and_delete_bucket(&self, name: &str, errors: &mut Vec<String>) {
```
Alphadelta14, line 1223: can we guarantee no further writes are going into this bucket? by revoking access etc

test/src/commands/purge.rs
```diff
@@ -0,0 +1,2244 @@
...
+                    .await,
+            );
+        }
+        Ok(errors)
+    }
+
+    /// Deletes the run's VPC(s).
+    async fn delete_vpcs(&self) -> Result<Vec<String>> {
+        let mut errors = Vec::new();
+        for vpc_id in self.vpc_ids().await? {
```
Alphadelta14, line 1613: if we pass in an existing vpc ever, this would be kinda scary. can we also check tags?

test/src/commands/purge.rs
```diff
@@ -0,0 +1,2244 @@
...
+            );
+        }
+        Ok(errors)
+    }
+
+    /// The serial network teardown spine: ENIs → security groups → subnets →
+    /// route tables → internet gateways → VPC.
+    async fn network_branch(&self) -> Result<Vec<String>> {
+        let mut errors = Vec::new();
+        collect(&mut errors, self.delete_network_interfaces().await);
```
Alphadelta14, line 1627: for my own edification, how parallelized are these async funcs?

Alphadelta14 (approved):
went through as thoroughly i could in a reasonable amount of time. there was a lot to it but things look logically alright

test/src/commands/purge.rs
alex-hunt-materialize: We've deleted node groups and instances before this. Karpenter runs on the node groups, so I'm not sure how a new instance would come up to attach the volume.

test/src/commands/purge.rs
alex-hunt-materialize: We delete the compute for everything that writes to it concurrently with the bucket cleanup. The bucket has always been deleted in the first pass in my testing, but if it happens to ever be blocked, it should get cleaned up in the next pass. I think that's still faster than blocking the bucket deletion waiting for the K8S cluster.

test/src/commands/purge.rs
alex-hunt-materialize: We have two main waves of concurrency, as defined in `sweep_once`. The first wave does everything except network and IAM concurrently, and the second wave does network and IAM concurrently.

Most operations within each category are serial within that category (ie: `network_branch` performs its operations serially). If we decide that's too slow, we can try running some of those concurrently too, but that was slightly more complicated and I was worried we might get rate limited.

test/src/commands/purge.rs
alex-hunt-materialize: `self.vpc_ids()` does check tags.

```rust
    /// Returns the VPC id(s) tagged for this run.
    async fn vpc_ids(&self) -> Result<Vec<String>> {
        let out = self
            .ec2
            .describe_vpcs()
            .filters(self.tag_filter())
            .send()
            .await
            .context("describe vpcs")?;
        Ok(out
            .vpcs()
            .iter()
            .filter_map(|v| v.vpc_id().map(String::from))
            .collect())
    }
```

test/src/commands/purge.rs
alex-hunt-materialize: I think this is the only place we check in this way, so moving it to a function doesn't get us much.

Alphadelta14: well, the others were slightly different, but the content shows up in quite a few places
```
➜  materialize-terraform-self-managed git:(purge) git grep -F '"TestRun"' test/src/commands/purge.rs | cat
test/src/commands/purge.rs:                .get("TestRun")
test/src/commands/purge.rs:                    .any(|t| t.key() == Some("TestRun") && t.value() == Some(&self.run_id));
test/src/commands/purge.rs:                    .any(|t| t.key() == Some("TestRun") && t.value() == Some(&self.run_id));
test/src/commands/purge.rs:                    .any(|t| t.key() == Some("TestRun") && t.value() == Some(&self.run_id));
test/src/commands/purge.rs:                    .any(|t| t.tag_key() == "TestRun" && t.tag_value() == self.run_id)
test/src/commands/purge.rs:                    (t.key() == Some("TestRun") && t.value() == Some(self.run_id.as_str()))
test/src/commands/purge.rs:                .any(|t| t.key() == "TestRun" && t.value() == self.run_id);
test/src/commands/purge.rs:                        .any(|t| t.key() == "TestRun" && t.value() == self.run_id),
test/src/commands/purge.rs:                    .any(|t| t.key() == "TestRun" && t.value() == self.run_id)
test/src/commands/purge.rs:                        .any(|t| t.key() == "TestRun" && t.value() == self.run_id),
test/src/commands/purge.rs:                    .any(|t| t.key() == "TestRun" && t.value() == self.run_id),
test/src/commands/purge.rs:                            .any(|t| t.key() == "TestRun" && t.value() == self.run_id)
```

PR #255, opened 2026-07-02
Parameterize GKE datapath provider (DEP-168)

alex-hunt-materialize:
## Summary

The `gcp/modules/gke` module hardcoded `datapath_provider = "ADVANCED_DATAPATH"` (Dataplane V2 / eBPF CNI). GKE cannot change the datapath provider on an existing cluster, so clusters built with the legacy CNI (everything created by the old terraform module) could not adopt this module without a full cluster rebuild.

- Add a `datapath_provider` variable to `gcp/modules/gke`, defaulting to `ADVANCED_DATAPATH` (no behavior change for current users). Accepts `ADVANCED_DATAPATH`, `LEGACY_DATAPATH`, and `DATAPATH_PROVIDER_UNSPECIFIED`.
- Expose the variable in the `simple` and `enterprise` GCP examples and pass it through to the gke module.
- Document that the legacy CNI silently ignores `NetworkPolicy` resources, and that changing the value forces cluster replacement.

Per discussion on the issue, this only parameterizes the CNI; network policy resources are still created by default (they are simply not enforced on the legacy datapath).

Refs [DEP-168](https://linear.app/materializeinc/issue/DEP-168/parameterize-gke-cni-and-make-network-policies-optional)

## Testing

Manually running the rust tests, once with the defaults and once with an overridden datapath_provider to use the legacy datapath.

### DIFF
--- a/gcp/examples/enterprise/main.tf
+++ b/gcp/examples/enterprise/main.tf
@@ -209,6 +209,7 @@ module "gke" {
   namespace                         = local.materialize_operator_namespace
   k8s_apiserver_authorized_networks = var.k8s_apiserver_authorized_networks
   labels                            = var.labels
+  datapath_provider                 = var.datapath_provider
 }
 
 # Create and configure generic node pool for all workloads except Materialize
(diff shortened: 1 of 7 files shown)


bosconi (approved):
Looks good.

PR #273, opened 2026-07-08
Expose polis_admin_login_emails on ory-stack

bobbyiliev:
The Polis admin UI is currently protected by a bootstrap SAML app (`tenant=_jackson_boxyhq`) that customers haven't configured yet, so it 403s. Plumbing `nextauth_acl` through as `polis_admin_login_emails` lets operators unlock the admin UI via email magic link and skip the API-only setup path.


kubernetes/modules/ory-stack/variables.tf
```diff
@@ -287,6 +287,13 @@ variable "polis_helm_values" {
   default     = {}
 }
 
+variable "polis_admin_login_emails" {
+  description = "Comma-separated allowlist of emails permitted to log into the Polis admin UI via magic link. Glob patterns are supported, so \"*@example.com\" allows any address in a domain. Empty is Polis's default and allows all authenticated users. Set this to unlock the admin UI without configuring a bootstrap SAML app; the alternative is driving Polis entirely through its admin API. Only meaningful when enable_polis is true."
+  type        = string
+  default     = ""
```
alex-hunt-materialize, line 293: Should we allow this default? Would it be better to default closed?

alex-hunt-materialize (dismissed):

kubernetes/modules/ory-stack/variables.tf
bobbyiliev: Yes that's fair, I just flipped it!

kubernetes/modules/ory-stack/main.tf
```diff
@@ -373,6 +373,8 @@ module "ory_polis" {
   admin_api_keys    = var.polis_admin_api_keys
   nextauth_secret   = var.polis_nextauth_secret
   db_encryption_key = var.polis_db_encryption_key
+  # Empty NEXTAUTH_ACL means allow-all in Polis; map to a sentinel so we ship closed.
+  nextauth_acl = var.polis_admin_login_emails != "" ? var.polis_admin_login_emails : "no-one@invalid.local"
```
alex-hunt-materialize, line 377: Does `null` also mean allow-all? Technically, `no-one@invalid.local` could be a real address, and it would be better to just disable this feature entirely.

kubernetes/modules/ory-stack/variables.tf
```diff
@@ -287,6 +287,13 @@ variable "polis_helm_values" {
   default     = {}
 }
 
+variable "polis_admin_login_emails" {
+  description = "Comma-separated allowlist of emails permitted to log into the Polis admin UI via magic link. Glob patterns are supported, so \"*@example.com\" allows any address in a domain. Empty (the default) locks the admin UI. Only meaningful when enable_polis is true."
+  type        = string
+  default     = ""
```
alex-hunt-materialize, line 293: Is there any scenario the customer won't be using this? Should we just make this a required field?

PR #274, opened 2026-07-08
Enterprise SSO end-to-end fixes and Okta group claims in the JWT

bobbyiliev:
A number of small fixes each in a separate commit. All of those are needed to make the Ory-based enterprise SSO stack more stable and better ux end-to-end. Verified with Okta SAML + SCIM.

- **LB outputs**: new `lb_addresses` output on `ory-stack`. Corrects `console_load_balancer_ip` on GCP and Azure enterprise examples, which was returning the operator's HTTP-only LB on port 8080 instead of the ory-stack HTTPS LB on 443.
- **Kratos**: adds the OIDC `session` hook on registration so first-time SAML users get a session cookie. Mapper now reads groups from `claims.raw_claims` too, and the identity schema includes a `groups` trait.
- **Polis**: overrides the chart's baked-in `OPENID_REDIRECT_EXACT_MATCH=true` back to origin-only matching so redirectUrl globs work.
- **OAuth2Client**: adds `postLogoutRedirectUris` so console logout stops 400-ing.
- **Selfservice UI**: initContainer patches `consent.js` at pod startup to propagate `identity.traits.groups` into the id_token / access_token. Temporary until [ory/kratos-selfservice-ui-node#389](https://github.com/ory/kratos-selfservice-ui-node/pull/389) lands upstream, then we drop the patch and set an env var.

Follow-ups: DEP-175 (operator support for suppressing the dead HTTP-only console LB).


kubernetes/modules/ory-selfservice-ui/main.tf
```diff
@@ -45,6 +45,23 @@ locals {
     COOKIE_SECRET      = local.cookie_secret
     CSRF_COOKIE_SECRET = local.csrf_cookie_secret
   }))
+
+  # Patches the upstream consent handler so identity.traits.groups is
```
bobbyiliev, line 49: This will be dropped once my upstream PR is merged: https://materializeinc.slack.com/archives/C0A0NEGLECV/p1783608539284379

azure/examples/enterprise/outputs.tf
```diff
@@ -139,15 +139,20 @@ output "materialize_instance_persist_backend_url" {
 
 # Load balancer outputs
 output "console_load_balancer_ip" {
-  description = "IP address of the Materialize console's load balancer."
-  value       = module.load_balancers.console_load_balancer_ip
+  description = "IP address of the browser-facing (HTTPS, port 443) Materialize console LoadBalancer, provisioned by the ory-stack module. Point the console DNS record at this address."
```
alex-hunt-materialize, line 142: It's a little weird that this load balancer is provisioned by the ory-stack module.

alex-hunt-materialize (approved):

PR #281, opened 2026-07-13
Retarget console LB to 443 for enterprise SSO instead of creating a duplicate LB

bobbyiliev:
Alex pointed out we already create a console LB in each cloud's LB module — this PR just retargets the existing one from 8080 to 443 (canonical HTTPS port, needed for Hydra's OAuth2 redirect URI and CORS to match) rather than adding a second LB. On Azure and GCP the existing `materialize_console_port` variable already supports this; on AWS the NLB submodule grows a small `console_listener_port` variable so the listener can bind to 443 while the target group keeps forwarding to the pod on 8080. The materialize -> ory egress NetworkPolicy also moves from ory-stack to materialize-instance (uses a `moved` block for in-place migration). Console LB IP will change on Azure/GCP during migration since it's now a different Service; enterprise SSO is still pre-GA so this is acceptable.


kubernetes/modules/materialize-instance/main.tf
```diff
@@ -247,3 +247,63 @@ resource "kubernetes_network_policy_v1" "allow_api_server_egress" {
     }
   }
 }
+
+# TCP-passthrough LoadBalancer on 443, TLS terminates at the console pod on 8080.
+resource "kubernetes_service_v1" "console_https_lb" {
```
alex-hunt-materialize: If the console is doing the TLS termination on 8080, why do we need this? Don't we already create a load balancer pointing at 8080 in `gcp/modules/load_balancers`, `azure/modules/load_balancers`, and `aws/modules/nlb`?

bobbyiliev: Oh, missed that entirely 🤦  you're right that's redundant. Thanks for catching it.

alex-hunt-materialize (approved):

PR #287, opened 2026-07-13
Batch renovate dependency updates

bobbyiliev:
Consolidates the open renovate dependency bumps (#284, #283) into a single PR and regenerates the module docs once with terraform-docs v0.24.0, so we run CI once instead of twice.

### DIFF
--- a/aws/modules/aws-lbc/versions.tf
+++ b/aws/modules/aws-lbc/versions.tf
@@ -16,7 +16,7 @@ terraform {
     }
     random = {
       source  = "hashicorp/random"
-      version = "~> 3.0, < 3.9.0"
+      version = "< 3.10.0"
     }
   }
 }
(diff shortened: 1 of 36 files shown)


aws/modules/aws-lbc/versions.tf
```diff
@@ -16,7 +16,7 @@ terraform {
     }
     random = {
       source  = "hashicorp/random"
-      version = "~> 3.0, < 3.9.0"
+      version = "< 3.10.0"
```
alex-hunt-materialize, line 19: Was this a manual change? This is a relaxation of the rules, which now allows downgrades, right?

We had discussed moving toward explicitly locking to specific versions.

alex-hunt-materialize (approved):

aws/modules/aws-lbc/versions.tf
bobbyiliev: Not manual, it's straight from renovate #283. But yea that's renovate rewriting the whole constraint. I'll try to open a PR to follow up to fix the renovate rangeStrategy.

PR #290, opened 2026-07-13
Stop renovate from dropping provider version floors

bobbyiliev:
Renovate's default replace strategy was collapsing compound constraints like `~> 3.0, < 3.3.0` down to just the upper bound (`< 3.4.0`), dropping the lower bound and allowing downgrades. This sets `rangeStrategy: widen` so future bumps keep the floor and only move the upper bound, and restores the floors that already got dropped across random, deepmerge, external, null, and azuread.

Verified with a renovate local dry-run: with the current config a `~> 3.0, < 3.5.0` constraint becomes `< 3.9.1` (floor dropped), while with `rangeStrategy: widen` it becomes `~> 3.0, < 3.9.1` (floor kept).

Docs regeneration will move to CI in a follow-up, since post-upgrade commands can't be allowlisted on our renovate offering.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,5 +2,11 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:recommended"
-  ]
+  ],
+  "rangeStrategy": "widen",
+  "postUpgradeTasks": {
+    "commands": ["bash .github/scripts/generate-docs.sh"],
+    "fileFilters": ["**/README.md"],
+    "executionMode": "branch"
+  }
 }
--- a/renovate.json
+++ b/renovate.json
@@ -3,10 +3,5 @@
   "extends": [
     "config:recommended"
   ],
-  "rangeStrategy": "widen",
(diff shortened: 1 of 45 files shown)


renovate.json
```diff
@@ -2,5 +2,11 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:recommended"
-  ]
+  ],
+  "rangeStrategy": "widen",
+  "postUpgradeTasks": {
+    "commands": ["bash .github/scripts/generate-docs.sh"],
```
Alphadelta14: our current offering has a very small set of tasks that are allowed to run, which we probably can't increase.

Alphadelta14 (dismissed):

renovate.json
bobbyiliev: Oh ok makes sense, I'll drop the `postUpgradeTasks` block and try to move the docs regeneration into CI instead.

Alphadelta14 (approved):

PR #300, opened 2026-07-14
Add node-local-dns caching on AWS and GCP

jubrad:
## What

Runs a per-node DNS cache on every node so pod DNS queries are answered on-node, without a network hop or a conntrack entry.

- **AWS**: new [`kubernetes/modules/node-local-dns`](../tree/node-local-dns/kubernetes/modules/node-local-dns) module wrapping the [deliveryhero node-local-dns helm chart](https://github.com/deliveryhero/helm-charts/tree/master/stable/node-local-dns) (pinned 2.1.0), wired into the simple and enterprise examples.
- **GCP**: GKE's built-in NodeLocal DNSCache addon (`dns_cache_config`), behind a new `enable_node_local_dns` variable on the gke module, enabled in the simple and enterprise examples.
- **Azure**: skipped, documented in `azure/README.md`.
- **Tests**: `verify` now waits for the `node-local-dns` DaemonSet rollout in kube-system on AWS and GCP.

## Why the per-cloud split

The helm chart intercepts the kube-dns ClusterIP with hostNetwork + NOTRACK iptables rules, which only works where services are resolved by kube-proxy in iptables mode (EKS). On eBPF dataplanes (GKE Dataplane V2, AKS with Cilium) the kube-dns ClusterIP is rewritten in eBPF before iptables sees it, so a self-deployed cache can never intercept:

- GKE: the built-in addon is the supported path.
- AKS: managed Cilium exposes no Local Redirect Policies. AKS's native [LocalDNS](https://learn.microsoft.com/en-us/azure/aks/localdns-custom) is the right path (node-local proxy, no interception needed) but is blocked on azurerm support for `localDNSProfile` ([azurerm#31342](https://github.com/hashicorp/terraform-provider-azurerm/issues/31342)); a local-exec workaround exists but any later azurerm node pool update could silently drop the profile and reimage the pool. Decision record in `azure/README.md`.

## Notable details

- **Taints/tolerations**: the chart's DaemonSet tolerates `CriticalAddonsOnly` plus all `NoSchedule`/`NoExecute` taints, so it runs on the tainted Materialize node pools, Karpenter nodes, and GCP swap-startup-tainted nodes with no extra config.
- **Cache TTLs**: the module overrides the chart's Corefile to cache cluster-zone records for 1s (chart default 30s). The custom CoreDNS deployment serves TTL-0 records so Materialize sees fresh pod IPs during rollouts; a 30s node-local cache would reintroduce that staleness. On GKE the addon's Corefile is not tunable and caches cluster records for ~5s (cache plugin minttl) even against TTL-0 records.
- **Bottlerocket**: the Corefile binds only the link-local IP and the kube-dns ClusterIP instead of 0.0.0.0. Wildcard :53 on the host network crashloops on Bottlerocket, which already has a listener on port 53 (bottlerocket-os/bottlerocket#3711). Found the hard way on a live test run; fix verified there (pods Running, NOTRACK rules installed, in-cluster lookups resolving).
- **GKE caveat**: toggling `dns_cache_config` on an existing cluster recreates all node pools; flagged in the variable description.

## Testing

- `terraform validate` on aws/gcp simple + enterprise examples; `terraform fmt`, `tflint`, terraform-docs regenerated with the CI-pinned v0.24.0.
- `cargo fmt` / `cargo clippy -D warnings` on the test harness.
- Live AWS test run (t260714-xz0bbg): DaemonSet rolled out on all 5 nodes, kube-dns VIP interception active, DNS resolution verified from a pod.


azure/README.md
```diff
@@ -98,3 +98,24 @@ provider "helm" {
...
 Your Azure project needs several APIs enabled. See the [examples/simple/README.md](./examples/simple/README.md#required-features) for the complete list of required APIs and how to enable them.
+
+### NodeLocal DNSCache
+
+The AWS and GCP examples run a node-local DNS cache on every node
+(the [`kubernetes/modules/node-local-dns`](../kubernetes/modules/node-local-dns)
+helm module on EKS, the built-in NodeLocal DNSCache addon on GKE). There is
+currently no equivalent here: these modules use Azure CNI powered by Cilium,
+whose eBPF dataplane resolves the kube-dns service IP before iptables runs, so
+a self-deployed node-local-dns can never intercept pod DNS traffic, and the
```
Alphadelta14: node-local-dns isn't supposed to "intercept" dns traffic. the resolvers for pods should be set to 169.254.x.y:53 instead of the cluster 10.x.y.10 kube-dns.kube-system.svc.
I'm fine with skipping AKS, just feels weird that this isn't supported.

kubernetes/modules/node-local-dns/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  # during rollouts, and the chart's default 30s node-local cache would
+  # reintroduce that staleness. __PILLAR__CLUSTER__DNS__ and
+  # __PILLAR__UPSTREAM__SERVERS__ are substituted by the node-cache binary at
+  # startup.
+  #
+  # The DaemonSet runs on the host network, so binding the wildcard address
+  # collides with anything else listening on port 53 on the node (Bottlerocket
+  # hosts already have a listener there, crashing the pod; see
+  # https://github.com/bottlerocket-os/bottlerocket/issues/3711). Bind only the
+  # link-local IP and the kube-dns ClusterIP, both of which node-cache sets up
```
Alphadelta14, line 27: oh, ew. we are binding the kube-dns ip?
This architecture goes very again how I'm usually used to seeing node-local-dns deployed. But it involves modifying the clusterDNS field of the ClusterConfig: https://docs.aws.amazon.com/eks/latest/eksctl/nodegroup-customize-dns.html.
I guess it makes it transparent, but it feels hacky.

Alphadelta14 (dismissed):
weird seeing it deployed this way, but I won't block on it.

kubernetes/modules/node-local-dns/main.tf
jubrad: Honestly, I was also really confused here as well... on the cloud side we handle this with local redirect policy, again v weird, clusterDNS per kubelet makes more sense to me, but this does seem like a standard deployment option.

Alphadelta14 (approved):

bobbyiliev (approved):

PR #301, opened 2026-07-15
modules: make local-exec scripts POSIX sh compatible

jasonhernandez:
## Motivation

The local-exec provisioners in eks-node-group, karpenter-nodepool, vpc-cni, and coredns hardcode `interpreter = ["/usr/bin/env", "bash", "-c"]` and use `set -euo pipefail`. That makes bash a hard runtime requirement wherever terraform runs these modules, which:

* blocks minimal container images — Materialize's BYOC `stack-deployer` wants to move to a distroless base whose only shell is busybox `sh`;
* is a general portability constraint for customers (macOS ships bash 3.2, some CI images have no bash at all).

Nothing in the scripts actually needs bash.

## Changes

* Drop the `interpreter` override — terraform's default is `/bin/sh -c`.
* `set -euo pipefail` → `set -eu`. Every pipeline in these scripts feeds `grep -q` inside a condition (or ends in the command whose failure matters, e.g. `... | xargs kubectl delete`), so dropping `pipefail` changes no behavior.
* Drop the bash-only `local` keyword in the ENI cleanup helper.
* A NOTE comment above each module's first provisioner records the POSIX-sh constraint so bashisms don't creep back in.

## Verification

* All 7 scripts extracted and checked: `shellcheck --shell=sh` clean at warning level except the pre-existing (and intentional) SC2064, and parse cleanly under `dash -n`.
* `terraform validate` passes for all four modules (1.15.2).
* Behavior is intentionally byte-identical aside from the shell used.

## Caveat

Destroy-time provisioners execute from state, so existing deployments keep running the old bash scripts until the `terraform_data` resources are next replaced. No action needed, just don't remove bash from an executor that still manages pre-change state.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -96,6 +96,58 @@ jobs:
             exit 1
           fi
 
+  shell:
+    name: Shell Lint
+    runs-on: ubuntu-latest
+    env:
+      SHELLCHECK_VERSION: "0.11.0"
+      SHELLCHECK_SHA256: "8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install ShellCheck
+        # Pinned by version and checksum, like the other tools here, so that a
+        # new shellcheck release cannot flip a green merge queue red without a
+        # code change in the PR.
+        run: |
+          set -euo pipefail
+          tarball="shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
+          curl -fsSL -o "$tarball" \
+            "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/${tarball}"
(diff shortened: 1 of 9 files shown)


.github/workflows/lint.yml
```diff
@@ -96,6 +96,58 @@ jobs:
...
+    name: Shell Lint
+    runs-on: ubuntu-latest
+    env:
+      SHELLCHECK_VERSION: "0.11.0"
+      SHELLCHECK_SHA256: "8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install ShellCheck
```
Alphadelta14, line 109: there are some shellcheck actions available directly.
https://github.com/reviewdog/action-shellcheck is reasonable and actually kept up to date.

Alphadelta14 (approved):
rip my poor favorite shell :'(

.github/workflows/lint.yml
jasonhernandez: not a lot of stars on that, so I'm leaning towards keeping the binary and hash right now

PR #311, opened 2026-07-20
ci: auto-regenerate module docs on Renovate PRs

jubrad:
## What

- `renovate.json`: label all Renovate PRs with `regenerate-docs`.
- `.github/workflows/renovate-docs.yml`: on PRs carrying that label, run `.github/scripts/generate-docs.sh` and commit the regenerated module READMEs back to the PR branch.

## Why

The `Check Generated Docs` job in `lint.yml` fails whenever a dependency bump changes module inputs/outputs. This keeps Renovate PRs green without manual `generate-docs.sh` runs.

## Notes

- Hosted Mend app cannot run `postUpgradeTasks`, so docs regenerate right after the PR opens rather than before. Label-driven trigger (not branch-name) so it also fires for any manually-labeled PR.
- `GITHUB_TOKEN` commits do not retrigger CI, so the push cannot loop.
- If `main` branch protection requires the `Check Generated Docs` status, confirm the bot's commit retriggers it (or the check may sit stale). Fork PRs can't be pushed to, but Renovate branches are in-repo.

### DIFF
--- a/.github/workflows/renovate-docs.yml
+++ b/.github/workflows/renovate-docs.yml
@@ -1,0 +1,58 @@
+name: Regenerate Docs
+
+# Regenerate module README docs on PRs carrying the "regenerate-docs" label
+# (Renovate applies it via renovate.json) and commit them back to the PR
+# branch. The lint.yml "Check Generated Docs" job fails when a dependency bump
+# changes module inputs/outputs; this keeps those PRs green without manual
+# intervention.
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+
+permissions:
+  contents: write
+
+jobs:
+  docs:
+    name: Regenerate Docs
+    # Gate on the label. The bot-token push DOES retrigger this workflow
+    # (synchronize), but the second run finds docs already current, so the
+    # "git diff --quiet" check no-ops and it cannot loop.
+    if: contains(github.event.pull_request.labels.*.name, 'regenerate-docs')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR branch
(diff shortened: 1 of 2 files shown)


bobbyiliev (dismissed):
I started on this in #297 but stalled on one thing I'm not 100% sure about, I think that pushes made with the default `GITHUB_TOKEN` don't retrigger CI (I think GitHub does that to avoid loops)?

So  `Check Generated Docs` fails, the regenerated commit might not flip it green. I didn't get to test it end to end, so might be worth checking.

Did you have a chance to test it?

I think we might end up reaching for a PAT / App token but never confirmed it fully.

.github/workflows/renovate-docs.yml
```diff
@@ -0,0 +1,54 @@
...
+# intervention.
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+    branches:
```
Alphadelta14: on.branches will run this on main too. you just want on.pull_request

.github/workflows/renovate-docs.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Generate docs
+        run: .github/scripts/generate-docs.sh
+
+      - name: Commit updated docs
+        run: |
+          if git diff --quiet; then
+            echo "Docs already up to date."
+            exit 0
+          fi
+          git config user.name "github-actions[bot]"
```
Alphadelta14, line 54: I added $MATERIALIZE_BOT_TOKEN to this repo if you need to push code that can rerun actions.
cc @bobbyiliev

Alphadelta14 (approved):

PR #318, opened 2026-07-22
feat(materialize-instance): enable RBAC by default (secure-by-default)

jasonhernandez:
## Summary

Adds an `enable_rbac` input to the `materialize-instance` module that sets
`spec.enableRbac` on the Materialize CR, **defaulting to `true`** so role-based
access control is enforced out of the box.

This is a **secure-by-default enhancement**, not a fix for any defect — the
module simply didn't expose the knob before.

## Why

Today the module never sets `spec.enableRbac`, so it falls to the CRD default
(`false`), and the operator then launches environmentd with
`--system-parameter-default=enable_rbac_checks=false`. The net effect is that a
default deployment does **not** enforce privilege checks unless the operator is
explicitly told otherwise. Enforcing RBAC by default is the safer posture for a
production-oriented module and matches the principle of least privilege.

## Change

- New `enable_rbac` variable (bool, default `true`).
- Wires `enableRbac = var.enable_rbac` into the CR spec.

```hcl
module "materialize_instance" {
  # ...
  enable_rbac = true   # default; set false to opt out
}
```

## Compatibility

This changes default behavior for deployments that relied on RBAC being off, so
it's worth calling out in release notes (opt out with `enable_rbac = false`). It
does not restrict the bootstrap admin — `mz_system` remains a superuser, so
existing admin automation continues to work; additional roles are simply subject
to privilege checks.


alex-hunt-materialize (changes requested):
The change looks good to me, but please add an upgrade note to the README.md in the root of the repo.

alex-hunt-materialize (approved):

bobbyiliev (approved):

PR #319, opened 2026-07-22
feat(gcp): blue-green node pool upgrades + node upgrade rollout trigger

alex-hunt-materialize:
## Motivation

GKE automatically upgrades node pools (this cannot be disabled), evicting Materialize pods and causing outages. Part of the [GCP nodegroup updates without major downtime](https://linear.app/materializeinc/project/gcp-nodegroup-updates-without-major-downtime-caba69cf4871/overview) Linear project, together with MaterializeInc/materialize#37807 (the orchestratord watcher).

Note that this will not take effect until we ship an orchestratord version that supports this.

## Changes

- **nodepool**: always use the autoscaled blue-green upgrade strategy — GKE creates the replacement (green) pool empty, cordons all original (blue) nodes, then waits `upgrade_wait_for_drain_duration` (default: the 3-day GKE default) before draining, with the cluster autoscaler scaling up the green pool as pods move. That wait is the window in which orchestratord rolls Materialize instances over gracefully. Requires the `google-beta` provider (the autoscaled rollout policy isn't in the GA provider) and a GKE control plane of 1.34.0-gke.2201000+ (STABLE channel is on 1.34.x).
- **gke**: publish `UpgradeEvent` cluster notifications to a new Pub/Sub topic (`enable_upgrade_notifications`, default on), create orchestratord's subscription, and grant its workload identity service account `roles/pubsub.subscriber` on it plus `roles/container.clusterViewer` for reading node pool upgrade state.
- **operator**: new `enable_node_upgrade_rollout_trigger` flag wiring the chart's `operator.cloudProvider.providers.gcp.nodeUpgradeRolloutTrigger` values (subscription, cluster name/location, watched node pools) plus operator service account annotations for workload identity. Also adds a network policy allowing operator egress to the GKE metadata server (`169.254.169.254:80`, plus `169.254.169.252:988` for the Dataplane V2 post-DNAT destination): workload identity credentials come over plain HTTP, which the existing egress policies (kube-system, 443, environmentd) blocked, leaving the watcher unable to authenticate ("no available authentication method found").
- **examples** (simple, enterprise): trigger enabled by default, watching the materialize node pool. The migration example gains the `google-beta` provider for the nodepool module (blue-green settings apply in-place, non-disruptively).

## Testing

Verified end-to-end on a test-harness run (us-east4, image `devel-b0b2754e` + local chart from the companion PR): watcher armed on the `UpgradeEvent`, polled until `WAITING_TO_DRAIN_BLUE_POOL` with the blue node cordoned, stamped `materialize.cloud/force-rollout` on the Materialize CR, and the new generation came up on the autoscaled green node with no evictions (`WaitUntilReady`).

## Notes for reviewer

- The trigger stays inert until an operator/chart release containing the watcher ships (examples pin `operator_version = v26.33.0`); the new helm values are ignored by older charts.
- The rollout only moves environmentd/clusterd; balancerd/console (and other pods) stay on the cordoned blue nodes until GKE's post-wait drain. Follow-up work (PDBs for envd/clusterd, `DISRUPTION_EVENT` handling) is planned separately.

### DIFF
--- a/gcp/examples/enterprise/main.tf
+++ b/gcp/examples/enterprise/main.tf
@@ -4,6 +4,14 @@ provider "google" {
   default_labels = var.labels
 }
 
+# Used by the nodepool module for autoscaled blue-green upgrade settings,
+# which are not yet available in the GA google provider.
+provider "google-beta" {
+  project        = var.project_id
+  region         = var.region
+  default_labels = var.labels
+}
+
 # Configure kubernetes provider with GKE cluster credentials
 data "google_client_config" "default" {}
 
@@ -217,6 +225,10 @@ module "gke" {
   # Cache DNS lookups on every node. On Dataplane V2 the GKE addon is the only
   # working node-local DNS option (see the variable description in the module).
   enable_node_local_dns = true
+
+  # Publish upgrade notifications to Pub/Sub for the node upgrade rollout
+  # trigger (see the operator module below).
+  enable_upgrade_notifications = var.enable_node_upgrade_rollout_trigger
 }
 
 # Create and configure generic node pool for all workloads except Materialize
@@ -387,6 +399,19 @@ module "operator" {
   # node selector for operator and metrics-server workloads
   operator_node_selector = local.generic_node_labels
 
+  # Trigger rollouts of Materialize instances when GKE upgrades the node
+  # pools they run on, so that their pods move to the replacement nodes
+  # gracefully instead of being evicted.
+  enable_node_upgrade_rollout_trigger    = var.enable_node_upgrade_rollout_trigger
+  node_upgrade_notification_subscription = module.gke.upgrade_notification_subscription
+  cluster_name                           = module.gke.cluster_name
+  cluster_location                       = module.gke.cluster_location
+  node_upgrade_watched_node_pools        = [module.materialize_nodepool.node_pool_name]
+  # Grant the operator workload identity access for its Pub/Sub subscription
+  # and GKE API reads.
+  operator_service_account_annotations = var.enable_node_upgrade_rollout_trigger ? {
+    "iam.gke.io/gcp-service-account" = module.gke.workload_identity_sa_email
+  } : {}
 
   # Enable Prometheus scrape annotations when observability is enabled
   helm_values = var.enable_observability ? {
--- a/gcp/examples/enterprise/main.tf
+++ b/gcp/examples/enterprise/main.tf
@@ -392,6 +392,10 @@ module "operator" {
   name_prefix = var.name_prefix
   region      = var.region
 
+  # Must match the namespace passed to the gke module: its workload identity
+  # binding targets the operator's service account in that namespace.
+  operator_namespace = local.materialize_operator_namespace
+
   # ARM tolerations and node selector for all operator workloads on GCP
   instance_pod_tolerations = local.materialize_tolerations
   instance_node_selector   = local.materialize_node_labels
@@ -408,7 +412,9 @@ module "operator" {
   cluster_location                       = module.gke.cluster_location
   node_upgrade_watched_node_pools        = [module.materialize_nodepool.node_pool_name]
   # Grant the operator workload identity access for its Pub/Sub subscription
-  # and GKE API reads.
+  # and GKE API reads. The gke module's workload identity binding targets the
+  # chart's service account (its orchestratord_service_account_name variable,
+  # default "orchestratord") in the operator namespace.
   operator_service_account_annotations = var.enable_node_upgrade_rollout_trigger ? {
     "iam.gke.io/gcp-service-account" = module.gke.workload_identity_sa_email
   } : {}
--- a/gcp/examples/simple/main.tf
+++ b/gcp/examples/simple/main.tf
@@ -348,6 +348,10 @@ module "operator" {
   name_prefix = var.name_prefix
   region      = var.region
 
+  # Must match the namespace passed to the gke module: its workload identity
+  # binding targets the operator's service account in that namespace.
+  operator_namespace = local.materialize_operator_namespace
+
   # ARM tolerations and node selector for all operator workloads on GCP
   instance_pod_tolerations = local.materialize_tolerations
   instance_node_selector   = local.materialize_node_labels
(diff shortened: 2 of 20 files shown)


gcp/modules/nodepool/variables.tf
```diff
@@ -154,3 +154,16 @@ variable "pause_container_resource_config" {
   }
   nullable = false
 }
+
+variable "upgrade_wait_for_drain_duration" {
+  description = "How long GKE waits after cordoning the blue pool before draining it during an (autoscaled blue-green) node pool upgrade. This is the window in which workloads can move to the replacement nodes gracefully. Between 0 and 7 days, as a duration in seconds (e.g. \"604800s\")."
+  type        = string
+  default     = "604800s" # 7 days, the maximum
```
bobbyiliev: Since orchestratord only rolls environmentd/clusterd, balancerd/console keep their blue nodes alive until the wait ends, right? With 7 days as default that overlap could be long, wondering if a shorter default is safer on cost.

gcp/modules/gke/variables.tf
```diff
@@ -136,3 +136,10 @@ variable "k8s_apiserver_authorized_networks" {
   }]
   nullable = false
 }
+
+variable "enable_upgrade_notifications" {
+  description = "Publish GKE upgrade notifications to a Pub/Sub topic and let orchestratord's service account subscribe to them and read node pool state. Required for the operator module's enable_node_upgrade_rollout_trigger."
+  type        = bool
+  default     = true
```
bobbyiliev, line 150: Small question: gke defaults this to `true` but the operator trigger defaults to `false`. Is it fine to create the Pub/Sub bits even when nothing uses them yet?

gcp/modules/gke/main.tf
```diff
@@ -109,6 +109,68 @@ resource "google_container_cluster" "primary" {
...
+  member       = "serviceAccount:${google_service_account.workload_identity_sa.email}"
+}
+
+# Lets orchestratord read node pool state (blue-green upgrade phase) to
+# decide when it is safe to trigger rollouts.
+resource "google_project_iam_member" "orchestratord_cluster_viewer" {
+  count = var.enable_upgrade_notifications ? 1 : 0
+
+  project = var.project_id
+  role    = "roles/container.clusterViewer"
```
bobbyiliev, line 172: This is project-level so the SA can list all clusters in the project, right? Low privilege, but wondering if scoping it down is worth it.

gcp/examples/simple/main.tf
```diff
@@ -343,6 +355,20 @@ module "operator" {
...
+  # pools they run on, so that their pods move to the replacement nodes
+  # gracefully instead of being evicted.
+  enable_node_upgrade_rollout_trigger    = var.enable_node_upgrade_rollout_trigger
+  node_upgrade_notification_subscription = module.gke.upgrade_notification_subscription
+  cluster_name                           = module.gke.cluster_name
+  cluster_location                       = module.gke.cluster_location
+  node_upgrade_watched_node_pools        = [module.materialize_nodepool.node_pool_name]
+  # Grant the operator workload identity access for its Pub/Sub subscription
+  # and GKE API reads.
+  operator_service_account_annotations = var.enable_node_upgrade_rollout_trigger ? {
```
bobbyiliev, line 374: This relies on the chart SA staying named `orchestratord` and the namespaces matching the gke module, if I read it right. Maybe worth a short note here?

gcp/examples/enterprise/main.tf
```diff
@@ -387,6 +399,19 @@ module "operator" {
...
+  # pools they run on, so that their pods move to the replacement nodes
+  # gracefully instead of being evicted.
+  enable_node_upgrade_rollout_trigger    = var.enable_node_upgrade_rollout_trigger
+  node_upgrade_notification_subscription = module.gke.upgrade_notification_subscription
+  cluster_name                           = module.gke.cluster_name
+  cluster_location                       = module.gke.cluster_location
+  node_upgrade_watched_node_pools        = [module.materialize_nodepool.node_pool_name]
+  # Grant the operator workload identity access for its Pub/Sub subscription
+  # and GKE API reads.
+  operator_service_account_annotations = var.enable_node_upgrade_rollout_trigger ? {
```
bobbyiliev, line 418: This relies on the chart SA staying named `orchestratord` and the namespaces matching the gke module, if I read it right. Maybe worth a short note here?

gcp/examples/simple/main.tf
```diff
@@ -343,6 +355,20 @@ module "operator" {
...
+    "iam.gke.io/gcp-service-account" = module.gke.workload_identity_sa_email
+  } : {}
+
```
bobbyiliev, line 377: Tiny nit: double blank line here, enterprise has one.

bobbyiliev (changes requested):
A few small questions and one nit, mostly to confirm intent. Nothing blocking on the approach itself.

gcp/modules/nodepool/variables.tf
alex-hunt-materialize: We can drop it back to the GCP default of 3 days. I increased it in case the rollout didn't work and they needed more time to debug it, but I guess a human can only intervene if they notice.

gcp/modules/gke/variables.tf
alex-hunt-materialize: I think it should be fine. There won't be many messages here, so the cost is negligible. I plan to swap the default for the operator when we release support for this there. I'd rather customers not have to touch their GKE module again.

gcp/modules/gke/main.tf
alex-hunt-materialize: We recommend that they run in a separate project just for us. We could put a condition on `resource.name`, which breaks `container.clusters.list` permission, or we could create a custom role. I don't think it's worth the complexity for what amounts to just listing clusters.

gcp/examples/enterprise/main.tf
alex-hunt-materialize: Even better, I can add a variable for it.

gcp/examples/simple/main.tf
alex-hunt-materialize: done

gcp/examples/enterprise/main.tf
alex-hunt-materialize: done

gcp/examples/simple/main.tf
alex-hunt-materialize: Added a variable

bobbyiliev (approved):

PR #320, opened 2026-07-22
ci: create releases from PR labels after merge

def-:
Motivation: https://materializeinc.slack.com/archives/C07PN7KSB0T/p1784207503.911529

The helm-chart publish pipeline in MaterializeInc/materialize currently pushes version bumps directly to `main` and blindly tags a patch release, without tests and regardless of what else is sitting on `main` unreleased. That can sweep breaking changes into a patch version. This PR moves tagging into this repository's CI instead:

* `pr.yml` gains a blocking `version-label` job that requires every PR to carry at least one label determining the version bump of the release created after merging: `breaking-change` (major), `enhancement` (minor), `bug`/`dependencies`/`documentation`/`github_actions` (patch), or `ignore-for-release` (no release). It feeds the existing required `ci-success` check, so no ruleset changes are needed.
* A new `tag.yml` runs on every push to `main`, computes the bump from the labels of the PRs merged since the latest tag (highest level wins), and creates the tag and GitHub release. Merges only happen through the merge queue, so the release is created after the infrastructure tests passed. A merge queue batch containing several PRs creates a single release.
* `renovate.json` additionally labels Renovate PRs with `dependencies` so they pass the new check and default to patch releases. Dependabot already labels its PRs with `dependencies` on its own. Humans can escalate a dependency PR to `breaking-change` before merging, like #296.

Notes:

* Direct pushes to `main` (bypassing the merge queue) are not tagged, their changes ship with the next labeled PR. That also makes the transition safe: the current publish pipeline's direct pushes have no associated PR, so `tag.yml` skips them and the pipeline's own tagging keeps working until a follow-up PR in MaterializeInc/materialize switches it to opening a labeled PR here.
* The currently unreleased #315 and #316 carry no labels, so merging this PR does not create a release. They will be part of the next labeled release.
* Manually pushed tags (e.g. for coordinated major releases) still work, the existing `release.yml` creates the GitHub release for those and `tag.yml` picks up the new tag as the baseline afterwards.
* The repository setting "Allow auto-merge" is currently disabled. Enabling it would let the publish pipeline mark its bump PRs as "merge when ready" so they merge and release right after a human approves them.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -23,6 +23,7 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          set -e \
           latest=$(gh api --paginate "repos/$GITHUB_REPOSITORY/tags" --jq '.[].name' \
             | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -n 1)
           echo "Latest release: $latest"
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,69 @@
+name: Tag
+
+on:
+  push:
+    branches: [main]
+
+permissions:
+  contents: write
+  pull-requests: read
+
+# Serialize runs so concurrent pushes cannot race on computing the next
+# version. A superseded pending run is skipped, its commits are picked up by
+# the newest run since every run considers everything since the latest tag.
+concurrency:
+  group: tag-main
+
+jobs:
+  tag:
+    name: Tag and Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create release for merged changes
+        env:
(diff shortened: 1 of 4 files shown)


Alphadelta14 (dismissed):
looks reasonable

.github/workflows/tag.yml
```diff
@@ -0,0 +1,69 @@
...
+jobs:
+  tag:
+    name: Tag and Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create release for merged changes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          latest=$(gh api --paginate "repos/$GITHUB_REPOSITORY/tags" --jq '.[].name' \
```
alex-hunt-materialize, line 27: Can we get a `set -e` at the top? There are a lot of network calls that could easily fail here.

alex-hunt-materialize (approved):

PR #333, opened 2026-07-29
Pull the Ory selfservice UI from the OEL registry proxy

bobbyiliev:
The selfservice UI was the one Ory component still on the community Docker Hub image (`oryd/kratos-selfservice-ui-node`), which ships an outdated base and a lot of unpatched CVEs. This adds imagePullSecret support to the module and points ory-stack at the OEL build behind the registry proxy, like Kratos/Hydra/Polis already are.

Opening as a draft because the exact OEL image path/tag still needs confirming with the registry team (marked with a TODO in the local).


bobbyiliev:
Closing this. The console team is porting the Ory login flow into the console itself using Ory elements, which deprecates the standalone Ory selfservice UI, so pulling its image through the OEL registry proxy is no longer needed.

PR #338, opened 2026-07-29
feat(ory-kratos): keep upstream OIDC client secrets out of the ConfigMap

jasonhernandez:
## Problem

The Ory Kratos Helm chart renders all of `kratos.config` into a **ConfigMap** (`templates/configmap-config.yaml`). Because this module passed `upstream_oidc_providers[].client_secret` through as part of that config, every upstream IdP client secret ended up in a ConfigMap, where it

- is **not** covered by Kubernetes Secret encryption at rest,
- is readable by anyone holding `get configmap` — a far more commonly granted verb than `get secret`,
- appears unredacted in `kubectl describe`, support bundles, and cluster backups.

The chart itself already treats `dsn`, `secrets.*`, and the SMTP URI as too sensitive for the ConfigMap and routes them through a Secret; upstream provider credentials were the remaining exception.

## Solution

The provider list — including client secrets — now lives in a Kubernetes Secret owned by the module and reaches Kratos as a single JSON-valued environment variable, injected via `secretKeyRef`:

```
SELFSERVICE_METHODS_OIDC_CONFIG_PROVIDERS = <JSON array of provider objects>
```

Kratos's config loader (`ory/x/configx`) JSON-decodes environment values into arrays and gives the variable precedence over the file configuration, so this works with every released Kratos. It is also the pattern ory/k8s#423 (the upstream ask for exactly this capability) was closed with.

**No provider credential reaches the Helm-rendered ConfigMap under any configuration** — there is deliberately no inline option and no new knob to hold wrong. `upstream_oidc_providers` keeps its interface; the delivery mechanism is entirely internal to the module.

## Implementation notes

- The rendered config keeps `selfservice.methods.oidc.enabled: true` with **no** `providers` key (schema-valid), so workloads that never receive the env var — the migration job, courier — boot unchanged.
- A `checksum/upstream-oidc-providers` pod-template annotation forces a rolling restart when the provider list changes: environment variables are immutable for running pods, and the chart's own checksum annotation only covers the ConfigMap.
- `deployment.extraEnv` and the TLS `extraVolumes` are assembled in one deployment block because mergo replaces lists rather than appending them.

## Migration

Existing deployments with `upstream_oidc_providers` set will see: one new Secret, the env var on the Kratos deployment, providers dropping out of the ConfigMap, and a single rolling restart. No variable changes are required.

Note (not written by Claude): given this feature is not yet released to customers, I think breaking changes are fine right now.

## Testing

- `terraform validate` on `ory-kratos` and `ory-stack`.
- Module READMEs regenerated with `.github/scripts/generate-docs.sh`.
- The env-var mechanism (JSON array decode, precedence over file config) verified against the Kratos codebase (`oryx/configx/koanf_env.go`) and empirically with a config-load test.

### DIFF
--- a/kubernetes/modules/ory-kratos/README.md
+++ b/kubernetes/modules/ory-kratos/README.md
@@ -35,6 +35,7 @@ No modules.
 | ---- | ---- |
 | [helm_release.kratos](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) | resource |
 | [kubernetes_namespace.kratos](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/namespace) | resource |
+| [kubernetes_secret.upstream_oidc_providers_env](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/secret) | resource |
 | [random_password.secrets_cipher](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) | resource |
 | [random_password.secrets_cookie](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) | resource |
 | [random_password.secrets_default](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) | resource |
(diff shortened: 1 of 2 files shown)


bobbyiliev (approved):
Looks good! Thanks @jasonhernandez!

PR #339, opened 2026-07-29
give the internal root ca a longer expiration

doy-materialize:
see https://github.com/MaterializeInc/cloud/pull/13068

### DIFF
--- a/kubernetes/modules/self-signed-cluster-issuer/main.tf
+++ b/kubernetes/modules/self-signed-cluster-issuer/main.tf
@@ -20,9 +20,11 @@ resource "kubectl_manifest" "self_signed_root_ca_certificate" {
       "namespace" = var.namespace
     }
     "spec" = {
-      "isCA"       = true
-      "commonName" = "${var.name_prefix}-self-signed-ca"
-      "secretName" = "${var.name_prefix}-root-ca"
+      "isCA"        = true
+      "commonName"  = "${var.name_prefix}-self-signed-ca"
+      "secretName"  = "${var.name_prefix}-root-ca"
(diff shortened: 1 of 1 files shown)


jubrad (dismissed):

kubernetes/modules/self-signed-cluster-issuer/main.tf
```diff
@@ -20,9 +20,11 @@ resource "kubectl_manifest" "self_signed_root_ca_certificate" {
...
     }
     "spec" = {
-      "isCA"       = true
-      "commonName" = "${var.name_prefix}-self-signed-ca"
-      "secretName" = "${var.name_prefix}-root-ca"
+      "isCA"        = true
+      "commonName"  = "${var.name_prefix}-self-signed-ca"
+      "secretName"  = "${var.name_prefix}-root-ca"
+      "duration"    = "8760h"
+      "renewBefore" = "2880h"
```
Alphadelta14, line 27: do we use this cert or do we have leaf certs we use? renewal of the root does not renew leaves (it only invalidates the existing leaves).
it's pretty common to just use 5y for a root in a self-signed limited context

doy-materialize: we use leaf certs, and yeah, we'll have to manually reach out to customers to get them renewed (https://materializeinc.slack.com/archives/CL68GT3AT/p1785360453550759)

alex-hunt-materialize (approved):

bobbyiliev (approved):

alex-hunt-materialize (approved):

PR #342, opened 2026-07-30
aws/database: add storage_type, iops, and storage_throughput variables

petrosagg:
Expose the RDS storage type and provisioned-performance knobs on the metadata-database module so users can select gp3 (or io1/io2) and provision IOPS/throughput, matching the metadata-database sizing guidance (which recommends gp3). All three default to null, preserving current behavior (the RDS module's provider default, gp2).


bobbyiliev (approved):
Looks good, thanks @petrosagg!

PR #357, opened 2026-08-04
MZC-53: Add per-service lb_overrides to the ory-stack module

jasonhernandez:
## Summary

Adds an `lb_overrides` input to `kubernetes/modules/ory-stack`: per-service overrides for the LoadBalancer Services, keyed by service role (`hydra`, `kratos`, `ui`, `polis` — the same keys as the `lb_addresses` output). Each entry can set:

- `annotations`, merged over the shared `lb_annotations` (override wins), so a deployment can mix internal and external LBs — e.g. internal Hydra/Kratos/selfservice UI with an internet-facing Polis for Okta SCIM provisioning.
- `source_ranges`, which sets `spec.loadBalancerSourceRanges`. The cloud controller enforces this in the provider firewall, so it restricts ingress even on clusters whose datapath does not enforce NetworkPolicy — on those clusters the existing `lb_source_cidrs` input is silently ignored (e.g. the legacy GKE datapath). The `lb_source_cidrs` description now documents that caveat.

First consumer is the context-graph staging environment (MZC-53), which will move Hydra/Kratos/selfservice UI to internal LBs reachable over its tailnet and restrict the Polis LB to Okta's published cell IP ranges, since SCIM is back-channel and needs inbound from Okta.

Part of MZC-53.

## Testing

- `terraform validate` on the module
- `terraform fmt` clean; README regenerated with `terraform-docs markdown table --hide-empty`
- No behavior change for existing callers: with `lb_overrides = {}` the rendered Services are identical


Alphadelta14 (approved):
looks right afaict

PR #359, opened 2026-08-05
DEP-184 New Observability Stack

Alphadelta14:
This replaces our weak grafana and prometheus installation with a full monitoring stack built for production environments:
* Thanos for long term metric storage
* Loki for logs
* Grafana with materialize dashboards
* Alloy Agent Daemonset collects logs (and soon cadvisor metrics) on each node and forwards them to the gateway
* Alloy Gateway processes metrics and logs and forwards them to long term destinations
  * Thanos and Loki are defaults, but other destinations such as GCM and Datadog are also supported
* alertmanager (no profiles just yet)
* kube-state-metrics, metrics-server

The same `enable_observability` flag today will enable these in the next major release.

## Structure
The main terraform for the materialize-monitoring module can be found in https://github.com/MaterializeInc/materialize-monitoring/tree/main/terraform/modules/materialize-monitoring documented in https://materializeinc.github.io/materialize-monitoring/getting-started/terraform/.
This had to live in alongside the helm chart to generally match the shape it was producing. However, that repo does not contain cloud-specific modules like this one does. So this repo just has a wrapper for each cloud setting up the cloud specific behavior.

The helm and TF modules are released in cadence here: https://github.com/MaterializeInc/materialize-monitoring/releases

### Escape hatches

We provide an arbitrary additional_values list(str) that can be passed to helm directly (usually as a yamlencoded document)

## Testing
Upstream runs two suites of E2E tests to validate the monitoring stack (without materialize). In the tier1 suite, we run using helm directly with local storage.
In the tier2 suite, we use rustfs for an S3 store and CNPG for postgresql with a "small" loki profile.
This helps stamp out most issues less cloud workload related issues (IAM, WIF, etc)

I've validated this in many helm scenarios under many different workload sizes. I've also gotten this up in GCP (with and without the optional GCM destination).

## Next Steps
This does not include setting up databases for Grafana yet nor does it include simple flags for other OTel destination (OTLP, Datadog).

## Other PRs
Docs: https://github.com/MaterializeInc/materialize/pull/38062
✅ Stack 1/3 - Set TF floor >= 1.10: https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/361
Stack 3/3 - https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/366 Add Grafana persistence and xLB

## For Reviewers
If you are overwhelmed by this changeset, I suggest focusing on just {aws,gcp,azure}/modules/monitoring/main.tf.
Pretty much everything else is superfluous to support those ~180 lines per cloud.


aws/modules/monitoring/main.tf
```diff
@@ -0,0 +1,320 @@
...
+
+  oidc_issuer_host = trimprefix(var.cluster_oidc_issuer_url, "https://")
+
+  common_tags = merge(var.tags, {
+    ManagedBy = "terraform"
+    Component = "materialize-monitoring"
+  })
+}
+
+resource "random_id" "bucket_suffix" {
```
alex-hunt-materialize, line 65: Instead of this, can we use account regional namespaced buckets? That way, we are guaranteed no collisions.

aws/modules/monitoring/main.tf
```diff
@@ -0,0 +1,320 @@
...
+  # (`../../../../materialize-monitoring/terraform/modules/materialize-monitoring`)
+  # for the duration. It must stay a *relative* path — an absolute one makes
+  # Terraform copy the module without the chart directory beside it, and the
+  # sizing profiles stop resolving.
+  #
+  # The tag's `/` is percent-encoded deliberately. Terraform 1.9 (which CI pins)
+  # splits the source on the first `/` inside the `ref` query value, silently
+  # truncating it to `?ref=materialize-monitoring` and failing the clone with
+  # `pathspec 'materialize-monitoring' did not match`. Newer Terraform parses it
+  # correctly, so this only reproduces on the pinned version.
```
alex-hunt-materialize: Can we bump the terraform version in CI and any minimum version specs? 1.9 is ancient.

azure/modules/monitoring/main.tf
```diff
@@ -0,0 +1,211 @@
...
+#     can read the other's data. That mirrors the AWS and GCP wrappers.
+#   * The webhook only mutates pods labelled `azure.workload.identity/use`, and
+#     the monitoring module applies that label — via `loki.podLabels` for Loki and
+#     `thanos.global.commonLabels` for Thanos, which has no `podLabels` of its
+#     own. The federated credentials below are what the projected tokens exchange
+#     against.
+#   * The `monitoring` namespace is created by the operator module in the
+#     supported topology, so `create_namespace` defaults to false. Keep a
+#     `depends_on` for the operator or the release can race the namespace.
+#   * Container lifecycle rules are off by default. Loki and Thanos both enforce
```
alex-hunt-materialize: Do we actually have lifecycle rules here in Azure?

azure/modules/monitoring/main.tf
```diff
@@ -0,0 +1,211 @@
...
+resource "random_string" "unique" {
+  length  = 6
+  special = false
+  upper   = false
+}
+
+# Account names are globally unique, lowercase alphanumeric, and capped at 24
+# characters — hence the suffix and the `replace`.
+resource "azurerm_storage_account" "telemetry" {
+  name                = substr(replace("${var.prefix}mzmon${random_string.unique.result}", "-", ""), 0, 24)
```
alex-hunt-materialize: Do we have a length max on the prefix, so this doesn't truncate the random portion?

gcp/modules/monitoring/main.tf
```diff
@@ -0,0 +1,269 @@
...
+# ==============================================================================
+# Workload Identity
+# ==============================================================================
+
+resource "google_service_account" "telemetry" {
+  for_each = local.service_accounts
+
+  # account_id is capped at 30 characters, so the prefix is truncated rather
+  # than allowed to overflow into an apply-time error.
+  account_id   = substr("${var.prefix}-mzmon-${each.key}", 0, 30)
```
alex-hunt-materialize: Truncation risk.

gcp/modules/monitoring/variables.tf
```diff
@@ -0,0 +1,182 @@
+variable "prefix" {
+  description = "Prefix for the bucket and service-account names this module creates."
+  type        = string
```
alex-hunt-materialize, line 3: Can we get validation that the prefix is under some length that fits without truncating? I think this applies to all clouds.

README.md
```diff
@@ -205,6 +205,27 @@ The GCP examples now default `region` to `us-east1` (previously `us-central1`),
 - To actually move an existing deployment to `us-east1`, treat it as a new deployment plus a data migration. GCP cannot relocate these resources in place, so there is no in-place `terraform apply` path between regions.
 - `gcp/README.md`'s `node_locations` examples now use `us-east1-b` and `us-east1-d`. `node_locations` must name zones inside the cluster's region. The module only regex-checks the `region-zone` string shape, so copies of the old `us-central1-*` examples pass `terraform validate` and `plan` and then fail at apply from the GKE API. Note `us-east1` has no `-a` zone.
 
+##### observability stack replaced (all clouds)
```
alex-hunt-materialize: We should probably say which version this ships in.

README.md
```diff
@@ -205,6 +205,27 @@ The GCP examples now default `region` to `us-east1` (previously `us-central1`),
...
+  - **GCP** — a GCS bucket and a Google service account per backend, bound with `roles/iam.workloadIdentityUser`. Requires Workload Identity on the cluster, which the `gke` module already sets.
+  - **Azure** — one storage account with a blob container per backend, a user-assigned managed identity per backend holding `Storage Blob Data Contributor` scoped to its own container, and a federated identity credential per ServiceAccount. Requires both `oidc_issuer_enabled` and `workload_identity_enabled` on the cluster, which the `aks` module already sets. The account is created with `shared_access_key_enabled = false`, so nothing falls back to a shared key.
+- Node pool capacity: the new stack runs microservice Loki, Thanos, Grafana, Alertmanager, kube-state-metrics, and two Alloy roles, against the previous stack's single Prometheus and Grafana. The `generic` pool may need to grow, or the first apply lands unschedulable pods.
+- If you set `install_metrics_server = false` on the operator module, set `install_metrics_server = true` on the monitoring module in the same change — the Materialize Console depends on the metrics API for cluster metrics.
```
alex-hunt-materialize: Does this have a possible race where they could delete/overwrite each other? Terraform doesn't have a way to know that they are the same kubernetes objects, right?

README.md
```diff
@@ -205,6 +205,27 @@ The GCP examples now default `region` to `us-east1` (previously `us-central1`),
...
+- **Azure only:** the Entra Workload ID webhook only mutates pods labelled `azure.workload.identity/use: "true"`, and the monitoring module applies that label for you. It reaches Thanos through `global.commonLabels` rather than a `podLabels` the Thanos chart does not have, so the label also appears on Thanos object metadata. That is cosmetic — it is not in any workload selector, so it is safe on an existing install.
+
+`enable_observability` keeps its name and its defaults (`false` in `simple`, `true` in `enterprise`).
```
alex-hunt-materialize: Can we toggle this to true in the tests, so that this code gets exercises by CI?

alex-hunt-materialize (commented):
Overall, this looks good.

aws/examples/simple/main.tf
```diff
@@ -553,14 +553,27 @@ module "materialize_instance" {
 }
 
 # 10. Setup Observability Stack (Prometheus + Grafana)
-module "prometheus" {
+module "monitoring" {
   count  = var.enable_observability ? 1 : 0
-  source = "../../../kubernetes/modules/prometheus"
+  source = "../../modules/monitoring"
```
bobbyiliev, line 558: Nice cleanup here. One thing I noticed: `bucket_force_destroy` is false by default, so once Loki and Thanos write anything, destroy would fail with BucketNotEmpty. We pass `bucket_force_destroy = true` to `module.storage` just above, want to do the same here?

aws/examples/enterprise/main.tf
```diff
@@ -539,14 +539,27 @@ module "materialize_instance" {
 }
 
 # 10. Setup Observability Stack (Prometheus + Grafana)
-module "prometheus" {
+module "monitoring" {
   count  = var.enable_observability ? 1 : 0
-  source = "../../../kubernetes/modules/prometheus"
+  source = "../../modules/monitoring"
```
bobbyiliev, line 544: Same thing as in the simple example, `bucket_force_destroy` would need to be true here or destroy gets stuck on the telemetry buckets.

gcp/examples/simple/main.tf
```diff
@@ -399,35 +422,34 @@ module "operator" {
   ]
 }
 
-module "prometheus" {
+module "monitoring" {
   count  = var.enable_observability ? 1 : 0
-  source = "../../../kubernetes/modules/prometheus"
+  source = "../../modules/monitoring"
```
bobbyiliev, line 433: Same thought as on the AWS side. `bucket_force_destroy` is false here while `gcp/modules/storage` hardcodes it to true, and GCS won't delete a non-empty bucket, so destroy would get stuck. Should we set it?

gcp/examples/enterprise/main.tf
```diff
@@ -454,33 +477,35 @@ module "operator" {
   ]
 }
 
-module "prometheus" {
+module "monitoring" {
   count  = var.enable_observability ? 1 : 0
-  source = "../../../kubernetes/modules/prometheus"
+  source = "../../modules/monitoring"
```
bobbyiliev, line 488: Same here for `bucket_force_destroy`.

aws/modules/monitoring/main.tf
```diff
@@ -0,0 +1,320 @@
...
+
+    loki_service_account_annotations = {
+      "eks.amazonaws.com/role-arn" = aws_iam_role.telemetry["loki"].arn
+    }
+    thanos_service_account_annotations = {
+      "eks.amazonaws.com/role-arn" = aws_iam_role.telemetry["thanos"].arn
+    }
+  }
+
+  additional_values = var.additional_values
```
bobbyiliev, line 361: Would it make sense to pass `chart_registry` and `enable_monitoring_crds` through too? I don't think either can be reached from `additional_values`, so folks on a mirrored registry or with existing prometheus-operator CRDs would have no way around it. Maybe `install_timeout` as well?

aws/modules/monitoring/variables.tf
```diff
@@ -0,0 +1,167 @@
...
+
+variable "node_selector" {
+  description = "Node selector for the centralized monitoring workloads. Not applied to the Alloy agent DaemonSet, which must reach every node."
+  type        = map(string)
+  default     = {}
+  nullable    = false
+}
+
+variable "storage_class" {
+  description = "StorageClass for the PVC-backed monitoring workloads (Alertmanager, the Loki ruler, and Thanos receive/compactor/store-gateway). Null uses the cluster default, which on EKS is `gp2`; the examples pass the `gp3` class from the `ebs-csi-driver` module."
```
bobbyiliev: Small doc thing, is gp2 still the EKS default? I think from 1.30 on EKS stopped marking it as default, and our own `ebs-csi-driver` module makes gp3 the default anyway. Maybe just "the cluster default class"?

gcp/modules/monitoring/main.tf
```diff
@@ -0,0 +1,269 @@
...
+  name          = each.value.name
+  location      = var.region
+  project       = var.project_id
+  force_destroy = var.bucket_force_destroy
+
+  # Required for IAM-only access; both backends authenticate as a service
+  # account, never with per-object ACLs.
+  uniform_bucket_level_access = true
+
+  versioning {
```
bobbyiliev, line 97: Something I ran into before: new GCS buckets get a 7 day soft delete policy by default and it does get billed. With versioning plus the 7 day noncurrent rule below, I think we'd pay twice for everything the compactors delete. Worth adding `soft_delete_policy { retention_duration_seconds = 0 }`?

aws/modules/monitoring/main.tf
```diff
@@ -0,0 +1,320 @@
...
+  bucket = each.value.id
+
+  rule {
+    apply_server_side_encryption_by_default {
+      sse_algorithm = "AES256"
+    }
+  }
+}
+
+resource "aws_s3_bucket_versioning" "telemetry" {
```
bobbyiliev, line 132: If I remember right, S3 versioning can only be suspended, never really turned off. So flipping `enable_bucket_versioning` to false later would leave the bucket versioned but drop the noncurrent expiry rule, and old versions would just keep growing. Would `Suspended` be safer than removing the resource?

gcp/modules/monitoring/main.tf
bobbyiliev: Good call adding the truncation, though I think with a prefix of 23+ chars both of these end up as the same `<prefix>-mzmon-`, so loki and thanos would collide and the id would end in a hyphen. Maybe trim the prefix instead, something like `"${substr(var.prefix, 0, 16)}-mzmon-${each.key}"`?

azure/modules/monitoring/main.tf
bobbyiliev: Same idea here I think. Past 13 chars the `substr` starts eating the random suffix, and past 19 there's none left, so two deployments with the same prefix would clash on a globally unique name. Could we trim the prefix and keep the random part whole?

azure/modules/monitoring/main.tf
```diff
@@ -0,0 +1,211 @@
...
+#     `thanos.global.commonLabels` for Thanos, which has no `podLabels` of its
+#     own. The federated credentials below are what the projected tokens exchange
+#     against.
+#   * The `monitoring` namespace is created by the operator module in the
+#     supported topology, so `create_namespace` defaults to false. Keep a
+#     `depends_on` for the operator or the release can race the namespace.
+#   * Container lifecycle rules are off by default. Loki and Thanos both enforce
+#     their own retention, and Thanos keeps blocks per downsampling resolution
+#     (raw / 5m / 1h) — a lifecycle rule deleting sooner removes blocks the
+#     compactor still references.
```
bobbyiliev: Might be missing it, but I don't see a way to actually turn lifecycle rules on in this module. AWS and GCP both have `logs_retention_days` / `metrics_retention_days` and `enable_bucket_versioning`, so should Azure get the same via `azurerm_storage_management_policy`, or should this bullet go for now?

aws/modules/monitoring/main.tf
```diff
@@ -0,0 +1,320 @@
...
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "telemetry" {
+  for_each = aws_s3_bucket.telemetry
+
+  bucket = each.value.id
+
+  rule {
```
bobbyiliev, line 118: Curious about this one: `aws/modules/storage` lets you pick SSE-KMS with your own key. Do we want the same option here, or is AES256 fine on purpose for telemetry?

gcp/examples/simple/main.tf
```diff
@@ -167,7 +167,30 @@ locals {
...
-  storage_class = "standard-rwo" # default storage class in gcp
+  # Not `standard-rwo`: the node pools are C4/C4A, which take only Hyperdisk —
+  # same constraint as their boot disks above. Every default GKE class is
+  # Persistent Disk, so a PVC on those pools never attaches.
+  storage_class = kubernetes_storage_class.hyperdisk_balanced.metadata[0].name
+}
+
+# GKE ships no Hyperdisk class, so create one. Not marked default — taking that
+# from `standard-rwo` would change provisioning for every other workload.
+resource "kubernetes_storage_class" "hyperdisk_balanced" {
```
bobbyiliev, line 184: Only the monitoring module reads `local.storage_class`, so this class shows up even with `enable_observability = false`. Is that intentional so other workloads can use it, or should it follow the flag?

gcp/examples/enterprise/main.tf
```diff
@@ -205,6 +208,26 @@ data "google_compute_zones" "available" {
   status  = "UP"
 }
 
+# GKE ships no Hyperdisk class, so create one. Not marked default — taking that
+# from `standard-rwo` would change provisioning for every other workload.
+resource "kubernetes_storage_class" "hyperdisk_balanced" {
```
bobbyiliev, line 219: Same question here about creating this even when observability is off.

azure/examples/simple/main.tf
```diff
@@ -102,6 +102,9 @@ locals {
   )
 
   materialize_instance_namespace = "materialize-environment"
+  # Matches the operator module's own default; named here so the monitoring
+  # module scopes its scrape targets to the same namespace.
+  materialize_operator_namespace = "materialize"
```
bobbyiliev, line 109: Just noticed `azure/modules/operator` takes `operator_namespace` and `monitoring_namespace`, but we don't pass either. Should this local go into `module.operator` too so they can't drift apart?

azure/examples/enterprise/main.tf
```diff
@@ -112,6 +112,9 @@ locals {
   )
 
   materialize_instance_namespace = "materialize-environment"
+  # Matches the operator module's own default; named here so the monitoring
+  # module scopes its scrape targets to the same namespace.
+  materialize_operator_namespace = "materialize"
```
bobbyiliev, line 119: Same here, worth wiring this local into `module.operator` as `operator_namespace`?

aws/examples/simple/main.tf
```diff
@@ -553,14 +553,27 @@ module "materialize_instance" {
 }
 
 # 10. Setup Observability Stack (Prometheus + Grafana)
-module "prometheus" {
+module "monitoring" {
```
bobbyiliev, line 556: Tiny one, the comment above still says Prometheus + Grafana. :)

aws/examples/enterprise/main.tf
```diff
@@ -539,14 +539,27 @@ module "materialize_instance" {
 }
 
 # 10. Setup Observability Stack (Prometheus + Grafana)
-module "prometheus" {
+module "monitoring" {
```
bobbyiliev, line 542: Same stale comment above here.

azure/modules/monitoring/versions.tf
```diff
@@ -0,0 +1,25 @@
+terraform {
+  # `optional()` in object type constraints requires 1.3.
+  required_version = ">= 1.3.0"
+
+  required_providers {
+    azurerm = {
+      source = "hashicorp/azurerm"
+      # Matches the constraint the other azure modules use, so a root that
+      # composes them resolves a single provider version.
+      version = ">= 3.75.0, < 4.76.0"
```
bobbyiliev: Should the floor be `>= 4.0` here? `storage_account_id` on `azurerm_storage_container` only landed in 4.x, so a 3.x provider wouldn't even parse it. I see `azure/modules/storage` has the same thing, so no big deal if you'd rather fix both separately.

.github/dependabot.yml
```diff
@@ -18,6 +18,7 @@ updates:
       - "/aws/modules/nlb"
       - "/aws/modules/nlb/target"
       - "/aws/modules/operator"
+      - "/aws/modules/monitoring"
```
bobbyiliev: Super minor, the list looks alphabetical so monitoring would go just before `/aws/modules/networking`. Same for azure and gcp below.

aws/modules/monitoring/variables.tf
```diff
@@ -0,0 +1,167 @@
...
+  nullable    = false
+}
+
+variable "cluster_oidc_issuer_url" {
+  description = "OIDC issuer URL of the cluster, from the EKS module."
+  type        = string
+  nullable    = false
+}
+
+variable "iam_permissions_boundary" {
```
bobbyiliev, line 48: Just curious, no other AWS module here takes a permissions boundary. Is someone already asking for it, or should we do it everywhere or nowhere?

bobbyiliev (dismissed):
This looks great overall, thanks for splitting the cloud wrappers out so cleanly, and for pinning the module and chart to one release so there is nothing to keep in sync.

Most of my notes are small. The two I'd like to sort out before merge are `bucket_force_destroy` in the examples (destroy will fail on the telemetry buckets as they are now) and passing `chart_registry` / `enable_monitoring_crds` through, since neither is reachable from `additional_values` and that blocks mirrored registries and clusters that already have the prometheus-operator CRDs. Everything else is a question or a nit, happy to be told I'm wrong on any of them.

One more general question on coverage: the simple examples default `enable_observability` to false and the e2e tests run those, so these modules only get `terraform validate` in CI and never an apply. Do we want to turn it on for one cloud, or is the upstream tier1/tier2 suite enough for now?

README.md
Alphadelta14: This is within the `#### v9.0.0` release notes

aws/modules/monitoring/main.tf
Alphadelta14: I bumped https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/361 conservatively.

README.md
Alphadelta14: helm knows and would conflict without adoption set. but they are in different namespaces. this is more of a TODO for us

README.md
Alphadelta14: I want this make this true by default! just anxious to in this changeset

azure/modules/monitoring/main.tf
Alphadelta14: not in azure. I don't see any issue with truncating the random portion though since it just decreases entropy.

aws/modules/monitoring/main.tf
Alphadelta14: not available in our current terraform-provider-aws 😩 
would require https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/288

Alphadelta14: which only makes it available. Filed DEP-201 for actually implementing the namespaced buckets (hopefully a fast-followup after resolving DEP-200-- the v6 upgrade)

gcp/examples/enterprise/main.tf
Alphadelta14: honestly it doesn't do any harm (doesn't actually allocate storage) and other modules/workloads may need it (we basically have this in the other clouds).

gcp/examples/simple/main.tf
Alphadelta14: called out in the other case: I think it's good to have a storageclass that actually works by default like our other clouds

aws/modules/monitoring/variables.tf
Alphadelta14: makes more sense when creating a role. we probably should do more of it.

alex-hunt-materialize: We support passing a permissions boundary to many AWS modules now, since it's required for BYOC.

eks
ebs-csi-driver
storage
vpc-cni
eks-node-group
aws-lbc
karpenter

alex-hunt-materialize (approved):

PR #361, opened 2026-08-06
Set floor of terraform >= 1.10; update CI to use latest TF

Alphadelta14:
This fixes a bug in 1.9 that our CI was running by simply changing the floor to a more recent version.
We recommend only the supported/stable versions of terraform (1.14 and 1.15) but it is easy to fall behind with a fast release cycle.
Our CI has moved to 1.15 and is now managed by renovate.

### DIFF
--- a/.github/workflows/test-aws.yml
+++ b/.github/workflows/test-aws.yml
@@ -76,7 +76,7 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4.0.1
         with:
-          terraform_version: "1.9.0"
+          terraform_version: "1.15.8"
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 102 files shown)


bobbyiliev (dismissed):

.github/workflows/test-aws.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4.0.1
         with:
-          terraform_version: "1.9.0"
+          terraform_version: "1.15.8"
```
alex-hunt-materialize, line 79: Should we test with the same version we require in our versions.tf files?

alex-hunt-materialize (approved):

.github/workflows/test-aws.yml
Alphadelta14: testing with oldest and newest would generally maximize coverage but that sounds terribly annoying.
I put into the notes that we should only be supporting 1.14 and 1.15 (the only [actively supported releases of terraform](https://endoflife.date/terraform)).
but 1.11 and 1.12 are still fairly around so forcing customers to upgrade terraform fully when things were working as is would be a bit of unnecessary friction.
especially since terraform is pretty loud about incompatibilities across versions.

PR #362, opened 2026-08-07
ory-stack: lock down Kratos self-service for SSO-only deployments

bobbyiliev:
Security fix for the SSO-only enterprise stack. Disables the Kratos password and profile self-service methods so users can only arrive through the upstream IdP and cannot self-register or edit their own identity traits. OIDC just-in-time provisioning is unaffected. Callers can override via `kratos_helm_values` if needed.


jubrad (approved):
This makes sense to me and we should get it in, but I do think we need to better outline what if any signups or user management we want outside of OIDC.

PR #364, opened 2026-08-07
ory-stack: make the standalone selfservice UI optional

bobbyiliev:
Adds `deploy_selfservice_ui` (default true, so behavior is unchanged). When set to false, the module skips the standalone selfservice UI deployment, its LoadBalancer, and its TLS cert, and points Kratos/Hydra at `selfservice_ui_url` instead. This lets another app (e.g. the console) host the Ory flow pages and drops one browser-facing endpoint.


jasonhernandez (approved):
thanks!

PR #366, opened 2026-08-07
CLO-111 DEP-202 DEP-196 Grafana Persistence and Load Balancers

Alphadelta14:
Stacked on https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/359. 
(Does not need to go into v9.0)

This adds Grafana persistence via dedicated database (uhg, RDS).
This also adds a load balancer like other exposed services.


alex-hunt-materialize (commented):
🤖 **Automated review by Claude** (run locally by @alex-hunt-materialize via Claude Code). Findings below were verified against the checked-out branch — `terraform validate` passes on all three monitoring modules and all six examples, and the pinned `materialize-monitoring` v0.15.0 module/chart was fetched to confirm value keys, variable names, and render-time checks. Please treat these as suggestions and use your judgement; a human should confirm before acting on any of them.

---

### 1. Public Grafana LB is now a hard plan failure on GCP (high)

`gcp/examples/enterprise/main.tf:526` (also `gcp/examples/simple/main.tf:470`)

`grafana_load_balancer` is wired unconditionally with `internal = var.internal_load_balancer` and `ingress_cidr_blocks = var.ingress_cidr_blocks`, whose default is `["0.0.0.0/0"]`. The module's third `grafana_load_balancer` validation refuses public + `0.0.0.0/0`, so `internal_load_balancer = false` — a documented, previously-working setting ("Set to `false` for prod-like demos validated against real DNS") — now fails `terraform plan` for anyone with `enable_observability = true`, which is the enterprise default.

The documented escape hatch is `additional_values = [yamlencode({connections = {grafana = {allowPublicAccess = true}}})]`, but no example exposes an `additional_values` root variable, and there is no `enable_grafana_load_balancer` toggle either. So the only ways out from inside the example are `enable_observability = false` or forking it.

Suggest either an `enable_grafana_load_balancer` toggle, or plumbing an `additional_values` root variable through the examples.

### 2. Same trap on Azure (high)

`azure/examples/enterprise/main.tf:489` (also `azure/examples/simple/main.tf:400`)

Identical to (1): `grafana_load_balancer` passed unconditionally, `ingress_cidr_blocks` defaults to `["0.0.0.0/0"]`, and `internal_load_balancer = false` becomes a plan error with no in-example acknowledgement path.

For contrast, `aws/examples/enterprise/main.tf:590` has the same shape but is gated on `grafana_host`, so it only affects users who explicitly opt into exposure. That gating seems like the right pattern to copy.

### 3. `ingress_cidr_blocks` validation crashes on null (medium)

`gcp/modules/monitoring/variables.tf:345` (identical expression at `azure/modules/monitoring/variables.tf:328`)

`length(var.grafana_load_balancer.ingress_cidr_blocks) > 0` and the `alltrue([for ...])` beside it both crash on a null list. The GCP examples explicitly permit `ingress_cidr_blocks = null` (`nullable = true`, and the root validation short-circuits on null), and `gcp/modules/load_balancers` documents it as *required* to be null when `internal = true`. So the natural internal-LB config dies with `Invalid function argument: ... argument must not be null` / `Iteration over null value` instead of the intended validation message. Reproduced against Terraform 1.15.2.

Suggested fix:

```hcl
var.grafana_load_balancer.ingress_cidr_blocks != null && length(var.grafana_load_balancer.ingress_cidr_blocks) > 0 && alltrue([for ...])
```

### 4. Grafana Cloud SQL instance is missing `random_instance_name` (medium)

`gcp/modules/monitoring/main.tf:341`

`module "grafana_database"` omits `random_instance_name = true`, so the Cloud SQL instance gets the deterministic name `<prefix>-mzmon-grafana-pg`. Both GCP examples set `random_instance_name = true` on `module.database` with an explicit comment, and the database module's own variable description notes that GCP reserves deleted instance names for ~1 week. A `terraform destroy` followed by `apply` inside that window — routine for the `simple` example — will fail with a 409 "instance already exists".

### 5. `tls = true` sets an https URL that nothing serves (low/medium)

`gcp/modules/monitoring/main.tf:397` (also `azure/modules/monitoring/main.tf:326`)

`tls = true` only flips `grafana_scheme`, producing `root_url = https://host` plus `grafana.ini security.cookie_secure = true`, while the chart's Grafana Service stays `port: 80 → targetPort: 3000` plaintext behind an L4 cloud LB. Nothing in the module terminates TLS or opens 443, so `https://host` cannot connect, and the secure cookie is dropped on the port that does — which is exactly the "nobody can log in" failure the adjacent comment warns about.

If the intent is "the operator terminates TLS themselves", the variable description should say so; right now the validation message ("the certificate both need the hostname") implies a certificate the module never creates.

---

<sub>Generated with [Claude Code](https://claude.com/claude-code)</sub>

alex-hunt-materialize (dismissed):

alex-hunt-materialize (approved):

PR #373, opened 2026-08-13
aws: upgrade to AWS provider v6 and EKS module v21

alex-hunt-materialize:
Migrates the AWS modules to `hashicorp/aws ~> 6.0` and `terraform-aws-modules/eks` v21 (DEP-200). Replaces renovate PR #288, whose widened constraint still resolved to 5.x.

v21 no longer bootstraps the self-managed addons, so everything clusters silently inherited from EKS is now managed explicitly: kube-proxy as an EKS addon, the aws-node service account by the VPC CNI chart, and the CoreDNS service account, RBAC, and kube-dns Service by the coredns module. The eks-node-group module pins the v20 launch-template defaults so upgrades do not roll nodes, and takes new `partition`/`account_id` inputs to keep IAM policy ARNs plan-time-known (without them, an upgrade plan silently detaches the node role's managed policies — see the README upgrade note).

Also carries a marked workaround for the upstream materialize-monitoring Loki bug (DEP-227).

Breaking release (v11.0.0) — upgrade steps and plan-review tripwires are in the README's upgrade notes.

**Testing** (manual, via the rust test helper; procedure in `test/aws-provider-v6-upgrade-test.md`):
- fresh install: apply + verify green
- v10 → v11 upgrade, with the new monitoring stack enabled: baseline on main, upgrade plan reviewed (saved in `test/upgrade-plan-*.txt`), apply + verify green, post-upgrade plan empty, node-role policies intact

### DIFF
--- a/test/Cargo.lock
+++ b/test/Cargo.lock
@@ -189,9 +189,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-cloudwatchlogs"
-version = "1.144.0"
+version = "1.145.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "56747e2f8f128b56d8befbe391592f4e6cc8f7a5ba9fe07134782cb42c3974b8"
+checksum = "84f68887a6c2bcb78e0329dbbefef18fa792031110796e90789f370cb777b6d5"
 dependencies = [
  "arc-swap",
  "aws-credential-types",
@@ -216,9 +216,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-ec2"
-version = "1.244.0"
+version = "1.246.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4eee9db6ca7ca0606330a5937105be43be98444f8b28ce93d9380f7506eea93b"
+checksum = "366e913d6ae975b6c0c79c61841026714ed5dd4e3284dac6b9ab00d9f06d96b2"
 dependencies = [
  "arc-swap",
  "aws-credential-types",
@@ -243,9 +243,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-eks"
-version = "1.140.0"
+version = "1.141.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "091c8a74c38f6de3d0eaf172a6577becc3f2db4efecc59820b1238df2c005f17"
+checksum = "083febe6f05648dd1da1ed3075788fac0377351164b97b691cd678aa228edd60"
 dependencies = [
  "arc-swap",
  "aws-credential-types",
@@ -322,9 +322,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-iam"
-version = "1.118.0"
+version = "1.119.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7707aab9b7951dbcdace83eb49893b05b2b7712fb9c1b9504592c95e04ec0999"
+checksum = "7469906ba3aa6de2b59dcb136543ad367b52eb135187c6733b9aa0dfa76982d9"
 dependencies = [
  "arc-swap",
  "aws-credential-types",
@@ -404,9 +404,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-s3"
-version = "1.140.0"
+version = "1.141.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e9660cf991e512fbe6094f1041ff3d3282bc5aeeeb6184e8de437ec2de024a10"
+checksum = "d9f9420d3a2467eed22ed3635ca653653162c386a0b0f65c78189f9bd3c1379e"
 dependencies = [
  "arc-swap",
  "aws-credential-types",
@@ -930,9 +930,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.6.5"
+version = "4.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "301b56658598e48f3648647ac6fc887be7e7108eddfa4e9b63fcf3ec58c0cadf"
+checksum = "473c7e07f409a8d772161724aa8db6a765a2532a70f9667eeb7b49d3d02fbdca"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -940,9 +940,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.6.5"
+version = "4.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "94a65403d1a1bd28f7dc68eb8506e8874808ee5eecb59298de588e2e1407a078"
+checksum = "7b48fea5a88e9ae728a2dcbedbfc0e730f7d60da42e1cb049a83c9fb8b789889"
 dependencies = [
  "anstream",
  "anstyle",
@@ -1404,9 +1404,9 @@ dependencies = [
 
 [[package]]
 name = "hcl-edit"
-version = "0.9.6"
+version = "0.9.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f641e65979f9c567246f13f9c64f7af41b3476e3171a63b10b36778af95b267b"
+checksum = "8a67cc5751cb5996669b9780cd738d4541a76c2d5f2f10c17af4965c5a03a5f5"
 dependencies = [
  "fnv",
  "hcl-primitives",
@@ -1417,9 +1417,9 @@ dependencies = [
 
 [[package]]
 name = "hcl-primitives"
-version = "0.1.11"
+version = "0.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "829a11d304c89e2cfe0dbb494a686bbe2b48ade17705c62cd1957b04aa4630f6"
+checksum = "bd662a8afeca01b5b5318f35baed70017b9f854bfa38bdcdadb87de946a49071"
 dependencies = [
  "itoa",
  "ryu",


test/upgrade-plan-t260810-jqhwjd.txt
```diff
@@ -0,0 +1,758 @@
+
```
bobbyiliev: Was this meant to be committed? It is a one off plan from run `t260810-jqhwjd`, it has a real account ID in it, and this repo is public. Maybe attach it to the PR description instead?

aws/modules/monitoring/main.tf
```diff
@@ -358,7 +358,32 @@ module "monitoring" {
...
 
-  additional_values = var.additional_values
+  # The first document works around a bug in the pinned materialize-monitoring
+  # module: on AWS it renders Loki's object_store without the S3 endpoint and
+  # region (it passes them to Thanos only), and Loki's thanos-objstore client
+  # requires an explicit endpoint, so every Loki component crash-loops at
+  # startup with "no s3 endpoint in config file". Remove once the upstream
+  # module renders these itself. Prepended so callers' additional_values can
+  # still override it.
+  additional_values = concat(
```
bobbyiliev: I think upstream already fixed this in v0.17.0, it derives the endpoint from the region now. #379 bumps to v0.17.0, so could this come back out when you rebase?

kubernetes/modules/coredns/main.tf
```diff
@@ -97,6 +101,46 @@ resource "kubernetes_config_map" "coredns" {
...
+    namespace = local.namespace
+    labels = {
+      "k8s-app"                       = "kube-dns"
+      "kubernetes.io/cluster-service" = "true"
+      "kubernetes.io/name"            = "CoreDNS"
+    }
+  }
+
+  spec {
+    cluster_ip = var.kube_dns_service_cluster_ip
```
bobbyiliev: If someone sets `create_kube_dns_service = true` and leaves this null, would Kubernetes just pick a random ClusterIP and break DNS? Maybe worth a validation to require it together?

aws/modules/networking/versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 5.0, < 5.101.0"
+      version = "~> 6.0"
```
bobbyiliev: The vpc module stays on `~> 5.0`. I ran validate on the enterprise example and it now warns about a deprecated `aws_region.name` from inside it. Worth bumping to v6 as well, or leave that for later?

bobbyiliev (commented):
Nice careful upgrade, just a few questions below, and I think it needs a rebase since it conflicts with main now.

Alphadelta14:
Verified this upgraded successfully in my AWS SM stack

Alphadelta14 (approved):

README.md
```diff
@@ -176,6 +176,43 @@ We follow semantic versioning with our tags. If a particular version requires ad
...
+   ```
+
+   This import path matches our examples. Your path may be different depending on where your coredns module is instantiated.
+
+   Skipping this fails at apply with `Service "kube-dns" already exists`.
+4. **Run `terraform plan`, review it against the notes below, then apply.**
+
+**Reviewing the plan:**
+
+- Expected changes: the `kube-proxy` EKS addon (adopts the existing self-managed kube-proxy), the CoreDNS service account and RBAC, a node security-group rule for port 10251, removal of the module's `terraform-aws-modules` tag and of a redundant cluster-encryption IAM policy (the KMS key policy retains the cluster grant), and in-place updates to the coredns and VPC CNI releases.
```
jubrad: **consider** — The expected-changes list is the tripwire an operator reads before applying, and it's missing two things your own captured plan shows.

First, `aws_iam_role_policy_attachment.this["AmazonEKSVPCResourceController"]` is destroyed — v21 dropped `enable_security_groups_for_pods`, so that policy comes off the cluster role. Most deployments here won't notice (it backs security groups for pods and Windows IPAM, not the CNI network-policy agent, which is node-agent based), but anyone running `SecurityGroupPolicy` workloads loses branch-ENI management with no error. Worth a line saying so, and pointing them at `iam_role_additional_policies` to put it back.

Second, the kube-proxy addon inherits v21's `most_recent = true` default, so it doesn't just adopt the existing daemonset — it also updates it to the newest version for the cluster's Kubernetes release and rolls kube-proxy on every node. That's a data-plane event during the upgrade window and belongs in the behavior notes even though no action is needed.

kubernetes/modules/coredns/main.tf
```diff
@@ -97,6 +101,46 @@ resource "kubernetes_config_map" "coredns" {
...
+    selector   = local.labels
+
+    port {
```
jubrad: **consider** — On the upgrade path this Service is imported over the EKS-bootstrapped one, and your captured plan shows the adoption is not a no-op:

```
- port {
    - name        = "metrics"
    - port        = 9153
  }
~ annotations = {
    - "prometheus.io/port"   = "9153" -> null
    - "prometheus.io/scrape" = "true" -> null
  }
```

The deployment still exposes `containerPort 9153` named `metrics`, so the port is real — it just stops being reachable through `kube-dns`. Anything scraping CoreDNS via that Service (annotation-based discovery, or a ServiceMonitor pointed at `kube-dns`) goes quiet without failing. Adding the third port and the two annotations here would make the import a true no-op and cost three lines; if you'd rather not carry them, the removal at least belongs in the plan-review notes.

test/upgrade-plan-t260810-jqhwjd.txt
```diff
@@ -0,0 +1,758 @@
...
+            "TestRun"     = "t260810-jqhwjd"
+            "deleteAfter" = "2026-08-14T13:54:39Z"
+            "owner"       = "Alex Hunt"
+            "reason"      = "materialize-terraform-self-managed integration test"
+            "team"        = "cloud"
+        }
+        # (10 unchanged attributes hidden)
+    }
+
+  # module.base_node_group.module.node_group.aws_iam_role_policy_attachment.this["AmazonEC2ContainerRegistryReadOnly"] must be replaced
```
jubrad: **consider** — This artifact was committed in `38b4c7d` (Aug 10), one day before `6e7a983` fixed the partition/account-id deferral — so the plan preserved in the repo is the *pre-fix* one, and it contains exactly the pattern the README tells operators to stop on:

```
# module.base_node_group.module.node_group.aws_iam_role_policy_attachment.this["AmazonEKS_CNI_Policy"] must be replaced
~ policy_arn = "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy" -> (known after apply) # forces replacement
```

Anyone who opens the file as "the reference plan for this upgrade" learns the wrong lesson. It also predates the vpc v6 bump, so it no longer represents the branch, and it carries a test account ID and cluster CA data into the repo for no ongoing benefit. I'd either regenerate it after the final config or drop it and keep the narrative in the README, which is where operators will actually look. (Related: `test/aws-provider-v6-upgrade-test.md`, referenced in the PR description as the procedure, isn't in the tree.)

aws/modules/eks-node-group/variables.tf
```diff
@@ -144,3 +144,17 @@ variable "iam_permissions_boundary" {
   type        = string
   default     = null
 }
+
+variable "partition" {
```
jubrad: **consider** — The failure mode when these are missing is the worst kind: the plan looks like a no-op replacement, the create is an AWS no-op, and the deposed destroy quietly detaches `AmazonEKSWorkerNodePolicy` and friends from a live node role. Every caller in this repo now passes them, but any consumer outside it (the BYOC stack-deployer, customer roots copied from an older example) that has a `depends_on` on this module call and doesn't update will hit it, and the only guard is a README they may not read.

Since this is already a breaking release, consider dropping the `""` defaults and making both required — the trap then surfaces as a plan-time "missing required argument" instead of a silent detach. If you'd rather keep them optional for callers with no `depends_on`, that's a reasonable call; I'd just make sure the BYOC side gets updated in lockstep.

kubernetes/modules/coredns/variables.tf
```diff
@@ -36,6 +36,19 @@ variable "create_coredns_service_account" {
...
 }
 
+variable "create_kube_dns_service" {
+  description = "Whether to create the kube-dns Service that exposes CoreDNS on the cluster DNS IP. Required on clusters whose platform does not bootstrap one (e.g. EKS clusters created with EKS module v21+). Clusters that already have a platform-bootstrapped kube-dns Service must import it into state instead: terraform import 'module.<name>.kubernetes_service.kube_dns[0]' kube-system/kube-dns"
+  type        = bool
+  default     = false
+  nullable    = false
+}
+
+variable "kube_dns_service_cluster_ip" {
```
jubrad: **consider** — `create_kube_dns_service = true` with `kube_dns_service_cluster_ip` left at `null` is silently catastrophic on a fresh cluster: Kubernetes assigns an arbitrary ClusterIP, kubelet's `--cluster-dns` still points at the `.10` address, and no pod resolves anything. The two variables are only ever meaningful together, so a `validation` block (or a `precondition` on the Service) turning that combination into a plan-time error would be worth the three lines — the module can't otherwise tell a caller they got this wrong.

aws/modules/vpc-cni/main.tf
```diff
@@ -59,17 +59,25 @@ resource "helm_release" "vpc_cni" {
...
-    value = "false"
+    value = "true"
   }
 
   set {
     name  = "serviceAccount.name"
     value = local.service_account_name
   }
 
+  set {
```
jubrad: **nit** — Now that the chart sets the IRSA annotation, `ROLE_ARN` in `terraform_data.annotate_existing_resources.input` (line 12) is dead — the script no longer reads it, and its header says as much. Worth removing so the provisioner's environment matches what the script documents as required.

jubrad (approved):
Careful migration — I checked the v21 renames, the count-gated `partition`/`account_id` data sources, the pinned launch-template defaults and the `bootstrap_self_managed_addons` ignore_changes against upstream 21.25.0 and they all hold up, and `terraform validate` passes on all three AWS examples; my comments are all around the edges of an otherwise sound change.

## Summary

5 considerations
1 nit

Alphadelta14:
gh stack cli closed this PR (it force pushed it to origin/main 😭).

PR #377, opened 2026-08-15
DEP-227 Enable observability by default; some fixes for AWS

Alphadelta14:
This enables observability by default. It is now in our tests and expected as a default path.

This adds the changes from https://github.com/MaterializeInc/materialize-monitoring/pull/231
> This fixes a few issues after doing things for real against AWS:
> 
> * S3 endpoint was broken in loki (thanos objstore client does not have a default S3 endpoint, so it needs an explicit one)
> * Alloy agent was not running on materialize workloads (tainted)-- we now support any NoSchedule like a typical monitoring daemonset
> * Relax HPA for alloy-gateway: do not consider memory when scaling (causes it to scale too much)
> * Add pre-delete hook for grafana operator resources to avoid a delete hang


aws/modules/monitoring/main.tf
```diff
@@ -403,8 +403,22 @@ locals {
...
+  # takes the short `name_prefix`, because it needs `create_before_destroy` and a
+  # fixed name would collide with itself on replacement.
+  #
+  # A trailing hyphen is invalid in a load-balancer name, so trim one if `substr`
+  # happens to land on it.
+  # Truncate the *prefix*, not the whole string, so the `-mzmon-grafana` half
+  # always survives — otherwise a long enough deployment prefix eats it and the
+  # load balancer is unidentifiable in the console, which is the thing a real
+  # `name` is here to avoid.
+  grafana_lb_name = "${trimsuffix(substr(var.name_prefix, 0, 32 - length("-mzmon-grafana")), "-")}-mzmon-grafana"
```
bobbyiliev: I think this can give two deployments the same name. `materialize-staging-blue` and `materialize-staging-green` both become `materialize-stagin-mzmon-grafana`, and load balancer names have to be unique per region per account.

`aws/modules/nlb` solves the same problem with a 6 character `name_prefix` plus an optional `nlb_name` to override it, and that way it can keep `create_before_destroy`. Could we do the same here?

README.md
```diff
@@ -182,7 +182,8 @@ Grafana gains durable state and a way to reach it. Both are opt-out rather than
 
 **Impact on existing deployments:**
 
-- Two new billable resources per deployment, **created whenever `enable_observability` is on**: the smallest PostgreSQL instance the cloud offers (`db.t4g.micro`, `db-f1-micro`, `B_Standard_B1ms`) and an internal L4 load balancer. On the `enterprise` examples, where observability defaults on, bumping `ref=<tag>` creates both. On `simple`, observability is off by default, so nothing new appears until you turn it on.
+- **`enable_observability` now defaults to `true` in the `simple` examples**, matching `enterprise`. The monitoring stack is opt-out rather than opt-in: bumping `ref=<tag>` on a `simple` root that never set the variable installs the whole stack — Loki, Thanos, Grafana, Alertmanager, kube-state-metrics, Alloy — along with its object storage and cloud identities. Set `enable_observability = false` to keep it off. The `generic` node pool may need to grow to fit it; see the v10.0.0 note below for what the stack runs.
```
bobbyiliev: `name` forces a new load balancer, so I think anyone already using the Grafana LB from v10.1.x gets a new NLB with a new DNS name after this. Should we add a bullet for it, so people know to repoint any DNS record or `grafana_host` they set?

aws/modules/monitoring/main.tf
```diff
@@ -450,7 +464,7 @@ locals {
 resource "aws_security_group" "grafana_nlb" {
   count = local.grafana_lb == null ? 0 : 1
 
-  name_prefix = "${local.grafana_lb_name_prefix}-mzmon-gf-"
+  name_prefix = "${var.name_prefix}-mzmon-grafana-"
```
bobbyiliev: Should the security group get `create_before_destroy` too? `aws/modules/database` sets it on its security group, and without it a later change that only replaces the security group would have to detach from the NLB first.

aws/modules/monitoring/main.tf
```diff
@@ -403,8 +403,22 @@ locals {
...
   grafana_lb = var.grafana_load_balancer
-  # `name_prefix` on an ELB is capped at 6 characters.
-  grafana_lb_name_prefix = substr(var.name_prefix, 0, min(6, length(var.name_prefix)))
+  # `aws_lb` and `aws_lb_target_group` cap `name_prefix` at 6 characters, and
+  # `name` at 32. The load balancer takes a full `name` — it is what shows in the
+  # console, so six characters of identity is not enough — and the target group
+  # takes the short `name_prefix`, because it needs `create_before_destroy` and a
+  # fixed name would collide with itself on replacement.
+  #
+  # A trailing hyphen is invalid in a load-balancer name, so trim one if `substr`
```
bobbyiliev: Small one, I think the name can never end with a hyphen, because it always ends with `-mzmon-grafana`. What `trimsuffix` really prevents is a double hyphen in the middle, when the prefix gets cut right on one. Maybe reword this?

aws/examples/simple/variables.tf
```diff
@@ -70,7 +70,7 @@ variable "internal_load_balancer" {
 variable "enable_observability" {
   description = "Enable Prometheus and Grafana monitoring stack for Materialize"
```
bobbyiliev: Since we are touching this anyway, should we fix the text? There is no Prometheus in the stack any more, it is Loki, Thanos, Grafana, Alertmanager and Alloy. Same description is in the other five examples.

bobbyiliev (commented):
Good catch on the `name_prefix` bug, that path could not have applied before with a prefix longer than 5 characters.

A few things below. The load balancer name is the main one, the rest are small.

aws/modules/monitoring/variables.tf
```diff
@@ -412,6 +412,20 @@ variable "grafana_service_port" {
...
 }
 
+variable "grafana_nlb_name" {
+  description = <<-EOT
+    Explicit name for the Grafana NLB. Null generates one from a six-character prefix plus the
+    suffix AWS appends, which is what keeps two deployments in one account and region from
+    colliding.
+
+    Set it when a predictable name is required. Note the tradeoff: a fixed name cannot have
+    `create_before_destroy`, so any change that replaces the load balancer costs a short outage
```
bobbyiliev: I might be reading this wrong, but `create_before_destroy` is set unconditionally on the `aws_lb`, and Terraform will not take a variable there. So with `grafana_nlb_name` set, would a replacement fail on a duplicate name instead of just costing a short outage? The same wording is in the README bullet.

bobbyiliev (dismissed):
Thanks for the fixes, this all looks good to me now.

aws/modules/monitoring/variables.tf
Alphadelta14: yeah, seem so. I don't think a destroy would be needed except for name changing though?

bobbyiliev (approved):

PR #379, opened 2026-08-17
DEP-204 Allow configuring OTLP / Datadog dests

Alphadelta14:
This adds support for OTLP destinations (like honeycomb) and Datadog. 

This works with the changes from https://github.com/MaterializeInc/materialize-monitoring/pull/247

---


jubrad (commented):
Clean pass-through — I checked the new variables against `materialize-monitoring` v0.17.0 and the names, object shapes, defaults and sensitivity all match, with validation left upstream where it belongs; my only real question is the Azure zone story, where the module docs and the examples disagree about what a non-zonal AKS pool looks like.

## Summary

3 considerations
1 nit

azure/modules/monitoring/variables.tf
```diff
@@ -165,6 +165,89 @@ variable "storage_class" {
...
 
+variable "min_zones" {
+  description = <<-EOT
+    Number of availability zones the node pool can actually launch in, used to adjust the hard zone
+    spread the chart puts on Thanos Receive and Loki's ingesters. Null leaves the chart's defaults
+    alone, which assume two or more zones.
+
+    Set it when that assumption does not hold, because the constraints fail closed rather than
+    degrading: `0` for a cluster whose nodes carry no `topology.kubernetes.io/zone` label, or the
+    real zone count otherwise. On AKS this is the one to check — an AKS node pool created without
```
jubrad: **consider** — This says an AKS node pool created without `zones` "is not zone-labelled" — but the examples set `min_zones = 1` and explain it as "a single domain to place into". By upstream's own semantics those are different cases: `0` is for a cluster whose nodes carry no `topology.kubernetes.io/zone` label at all (the hard constraints get dropped), `1` is for one real zone (`minDomains: 1`). If the nodes truly carry no zone label, then `minDomains: 1` on a `topologyKey` no node has still leaves the pods Pending forever, which is exactly the failure this knob exists to avoid.

My read is that AKS labels non-zonal nodes `topology.kubernetes.io/zone=0`, so the examples are right and this sentence is what needs fixing — but since it's a fail-closed constraint, worth confirming against a real non-zonal cluster (`kubectl get nodes -L topology.kubernetes.io/zone`) and then making the two texts agree.

gcp/modules/monitoring/main.tf
```diff
@@ -322,6 +322,17 @@ module "monitoring" {
     prefix         = var.google_cloud_metrics_prefix
   } : null
 
+  # Straight pass-through; the monitoring module validates the tiers, the OTLP
+  # protocol, and the url-has-no-scheme rule, and rejects a bearer token combined
+  # with headers. Credentials never enter the Helm values — that module puts them
+  # in the gateway Secret and rolls the gateway when one changes.
+  datadog_metrics = var.datadog_metrics
```
jubrad: **consider** — `min_zones` is propagated on AWS and Azure but not here. That's defensible for the examples — the GKE module creates a regional cluster (`location = var.region`), so the chart's `minDomains: 2` default holds — but `node_locations` is a caller input, and anyone who pins it to a single zone hits the same Pending-forever behaviour with no way out of this module except `additional_values`.

Since the variable is already written twice, a third copy is cheap and keeps the three wrappers reading the same. If it's deliberately out of scope, that's fine too — just worth being a decision rather than an omission.

aws/examples/simple/main.tf
```diff
@@ -576,6 +576,22 @@ module "monitoring" {
...
   storage_class = module.ebs_csi_driver.storage_class_name
 
+  # Datadog and generic OTLP (Honeycomb, Grafana Cloud, your own collector) fan
+  # out the same way, and need no cloud resources — so they are set here rather
+  # than behind an `enable_*` toggle. Commented out because both need a
+  # credential; the module puts it in a Secret rather than the Helm values, and
+  # rolls the gateway when it changes.
+  #
+  # datadog_metrics = { site = "datadoghq.com", min_importance = "essential" }
+  # datadog_api_key = var.datadog_api_key
```
jubrad: **consider** — Two things about this block being commented out.

First, as written it can't be uncommented: `var.datadog_api_key` and `var.honeycomb_api_key` aren't declared in this example's `variables.tf`, and the same holds in the other five examples.

Second — and this is the actual suggestion — I think these want to be real variables rather than comments. The module inputs are all null-defaulted no-ops, so wiring them unconditionally costs nothing when unset:

```hcl
datadog_metrics = var.datadog_metrics
datadog_api_key = var.datadog_api_key

otlp_metrics             = var.otlp_metrics
otlp_auth_header_secrets = var.otlp_auth_header_secrets
```

with matching null-default (and `sensitive`, for the credentials) variables in `variables.tf`. That's the pattern the rest of these examples already use for optional inputs — `grafana_host`, `materialize_version`, `enable_observability` — and it means a credential arrives through `terraform.tfvars` or `TF_VAR_`, which is where it belongs. The commented shape quietly steers a reader toward pasting the key into `main.tf` instead, and a Datadog key hardcoded in a committed example is a worse failure mode than a verbose `variables.tf`.

The cost is real: mirroring the two object types across six example directories is about thirty lines each of pure pass-through. If that's not worth it, the cheap version is to keep the block commented but make it self-contained — literal placeholders instead of `var.` references that don't resolve, and a note that the credential should come from a variable rather than a literal.

aws/modules/monitoring/variables.tf
```diff
@@ -181,6 +181,89 @@ variable "storage_class" {
...
+
+variable "otlp_auth_header_secrets" {
+  description = "Secret request headers for `otlp_metrics`, as header name to value — Honeycomb's `x-honeycomb-team`, for instance. Each reaches the gateway as a Secret the monitoring module creates. Cannot be combined with `otlp_auth_bearer_token`."
+  type        = map(string)
+  default     = {}
+  nullable    = false
+  sensitive   = true
+}
+
+variable "otlp_auth_bearer_token" {
```
jubrad: **nit** — Upstream also rejects `otlp_auth_bearer_token` combined with `otlp_metrics.auth_headers`, not just with `otlp_auth_header_secrets` — the chart has one auth slot per destination and non-secret inline headers land in the same list. As written, a reader could reasonably set a bearer token plus `auth_headers = { "x-honeycomb-dataset" = ... }` and get a plan failure this description told them wouldn't happen. Worth naming both here (and in the two copies).

aws/examples/simple/main.tf
Alphadelta14: should we be exposing everything as variables? I don't want to be making the space a bit too noisy.

aws/modules/monitoring/variables.tf
Alphadelta14: this rejects pretty early and I don't want to add the validation at every layer, so I think it's fine

azure/modules/monitoring/variables.tf
Alphadelta14: yeah, it does get the zone label. just not with a meaningful value always (like `topology.kubernetes.io/zone: 0`). will update comment.

gcp/modules/monitoring/main.tf
Alphadelta14: this was a hiccup on my rebase since the min_zones actually went into the second PR. I'm new to managing stacks via the gh cli and got confused 😅

aws/examples/simple/main.tf
jubrad: I think the general pattern in this repo to expose minimal variables... personally... I prefer reasonable defaults and tons of levers, but it's a good point and perhaps keeping these as non vars is right?

azure/examples/enterprise/main.tf
```diff
@@ -469,6 +469,30 @@ module "monitoring" {
   node_selector = local.generic_node_labels
   storage_class = local.storage_class
 
+  # The AKS node pools here are non-zonal — `availability_zones` is left at its
```
bobbyiliev: I think this comment was copied over from the simple example. Here we do set `availability_zones = ["1", "2", "3"]` on line 61, and it goes to the default node pool, which is where monitoring runs because of `node_selector`. Should we drop `min_zones` so it stays at the chart's `minDomains: 2`, or set it to 3 to match the zones we have?

gcp/modules/monitoring/variables.tf
```diff
@@ -148,8 +148,10 @@ variable "node_selector" {
 
 variable "storage_class" {
   description = <<-EOT
-    StorageClass for the PVC-backed monitoring workloads (Alertmanager, the Loki ruler, and Thanos
-    receive/compactor/store-gateway). Null uses the cluster default.
+    StorageClass for the PVC-backed monitoring workloads (Alertmanager, the Loki ruler, and the
+    Thanos Store Gateway and Compactor). Null uses the cluster default. Loki's ingesters and Thanos
+    Receive use node-local `emptyDir` by design, so the class only reaches them if their persistence
```
bobbyiliev: Small one, I think this part is only true for Thanos Receive. Upstream `storage_class.tf` does set a `thanos.receive` key, but it leaves out `loki.ingester` on purpose, so turning Loki ingester persistence back on would not pick up the class. Maybe split the sentence in two?

Same sentence is in `aws/modules/monitoring/variables.tf` and `azure/modules/monitoring/variables.tf`.

gcp/modules/monitoring/variables.tf
```diff
@@ -183,6 +200,73 @@ variable "google_cloud_metrics_prefix" {
...
+variable "otlp_auth_header_secrets" {
+  description = "Secret request headers for `otlp_metrics`, as header name to value — Honeycomb's `x-honeycomb-team`, for instance. Each reaches the gateway as a Secret the monitoring module creates. Cannot be combined with `otlp_auth_bearer_token`."
+  type        = map(string)
+  default     = {}
+  nullable    = false
+  sensitive   = true
+}
+
+variable "otlp_auth_bearer_token" {
+  description = "Bearer token for `otlp_metrics`, for endpoints taking `Authorization: Bearer`. Reaches the gateway as a Secret the monitoring module creates. Cannot be combined with `otlp_auth_header_secrets`."
```
bobbyiliev: I think upstream also blocks a bearer token together with `otlp_metrics.auth_headers`, not only `otlp_auth_header_secrets`. The example above uses `auth_headers` for the dataset, so it would be easy to hit this by accident. Can we mention both here?

Same description is in the aws and azure modules.

aws/examples/simple/main.tf
bobbyiliev: Tiny thing, `var.datadog_api_key` and `var.honeycomb_api_key` are not declared in this example, so anyone who uncomments this would need to add them first. Maybe worth a short note in the comment?

Same block is in the other five examples.

bobbyiliev (changes requested):
Just a few small things.

aws/modules/monitoring/variables.tf
Alphadelta14: oh this was more of a comment state thing. you right; i read that wrong. added more context to this one

azure/modules/monitoring/variables.tf
```diff
@@ -160,11 +160,100 @@ variable "node_selector" {
...
 
+variable "min_zones" {
+  description = <<-EOT
+    Number of availability zones the node pool can actually launch in, used to adjust the hard zone
+    spread the chart puts on Thanos Receive and Loki's ingesters. Null leaves the chart's defaults
+    alone, which assume two or more zones.
+
+    On AKS if you end up with a single zone (which is how node pools in AKS are created by default
+    without `zones`), you MUST set this to `1` or the pods will remain **Pending** forever.
+    Set it to `0` to disable the zone spread entirely (only required if there is no
```
jubrad: **consider** — This says an AKS node pool created without `zones` "is not zone-labelled" — but the examples set `min_zones = 1` and explain it as "a single domain to place into". By upstream's own semantics those are different cases: `0` is for a cluster whose nodes carry no `topology.kubernetes.io/zone` label at all (the hard constraints get dropped), `1` is for one real zone (`minDomains: 1`). If the nodes truly carry no zone label, then `minDomains: 1` on a `topologyKey` no node has still leaves the pods Pending forever, which is exactly the failure this knob exists to avoid.

My read is that AKS labels non-zonal nodes `topology.kubernetes.io/zone=0`, so the examples are right and this sentence is what needs fixing — but since it's a fail-closed constraint, worth confirming against a real non-zonal cluster (`kubectl get nodes -L topology.kubernetes.io/zone`) and then making the two texts agree.

gcp/modules/monitoring/main.tf
```diff
@@ -322,6 +323,17 @@ module "monitoring" {
     prefix         = var.google_cloud_metrics_prefix
   } : null
 
+  # Straight pass-through; the monitoring module validates the tiers, the OTLP
+  # protocol, and the url-has-no-scheme rule, and rejects a bearer token combined
+  # with headers. Credentials never enter the Helm values — that module puts them
+  # in the gateway Secret and rolls the gateway when one changes.
```
jubrad: **consider** — `min_zones` is propagated on AWS and Azure but not here. That's defensible for the examples — the GKE module creates a regional cluster (`location = var.region`), so the chart's `minDomains: 2` default holds — but `node_locations` is a caller input, and anyone who pins it to a single zone hits the same Pending-forever behaviour with no way out of this module except `additional_values`.

Since the variable is already written twice, a third copy is cheap and keeps the three wrappers reading the same. If it's deliberately out of scope, that's fine too — just worth being a decision rather than an omission.

aws/examples/simple/main.tf
```diff
@@ -576,6 +576,25 @@ module "monitoring" {
...
   storage_class = module.ebs_csi_driver.storage_class_name
 
+  # Datadog and generic OTLP (Honeycomb, Grafana Cloud, your own collector) fan
+  # out the same way, and need no cloud resources — so they are set here rather
+  # than behind an `enable_*` toggle. Commented out because both need a
+  # credential; the module puts it in a Secret rather than the Helm values, and
+  # rolls the gateway when it changes.
+  #
+  # Declare the two credentials as `sensitive` variables of your own before
+  # uncommenting — this example does not, and they belong in `terraform.tfvars`
```
jubrad: **consider** — Two things about this block being commented out.

First, as written it can't be uncommented: `var.datadog_api_key` and `var.honeycomb_api_key` aren't declared in this example's `variables.tf`, and the same holds in the other five examples.

Second — and this is the actual suggestion — I think these want to be real variables rather than comments. The module inputs are all null-defaulted no-ops, so wiring them unconditionally costs nothing when unset:

```hcl
datadog_metrics = var.datadog_metrics
datadog_api_key = var.datadog_api_key

otlp_metrics             = var.otlp_metrics
otlp_auth_header_secrets = var.otlp_auth_header_secrets
```

with matching null-default (and `sensitive`, for the credentials) variables in `variables.tf`. That's the pattern the rest of these examples already use for optional inputs — `grafana_host`, `materialize_version`, `enable_observability` — and it means a credential arrives through `terraform.tfvars` or `TF_VAR_`, which is where it belongs. The commented shape quietly steers a reader toward pasting the key into `main.tf` instead, and a Datadog key hardcoded in a committed example is a worse failure mode than a verbose `variables.tf`.

The cost is real: mirroring the two object types across six example directories is about thirty lines each of pure pass-through. If that's not worth it, the cheap version is to keep the block commented but make it self-contained — literal placeholders instead of `var.` references that don't resolve, and a note that the credential should come from a variable rather than a literal.

aws/modules/monitoring/variables.tf
```diff
@@ -176,11 +176,101 @@ variable "node_selector" {
...
+    protocol       = optional(string, "grpc")
+    compression    = optional(string)
+    min_importance = optional(string, "recommended")
+    auth_headers   = optional(map(string), {})
+  })
+  default = null
+}
+
+variable "otlp_auth_header_secrets" {
+  description = "Secret request headers for `otlp_metrics`, as header name to value — Honeycomb's `x-honeycomb-team`, for instance. Each reaches the gateway as a Secret the monitoring module creates. Cannot be combined with `otlp_auth_bearer_token`."
```
jubrad: **nit** — Upstream also rejects `otlp_auth_bearer_token` combined with `otlp_metrics.auth_headers`, not just with `otlp_auth_header_secrets` — the chart has one auth slot per destination and non-secret inline headers land in the same list. As written, a reader could reasonably set a bearer token plus `auth_headers = { "x-honeycomb-dataset" = ... }` and get a plan failure this description told them wouldn't happen. Worth naming both here (and in the two copies).

jubrad (approved):
Clean pass-through — I checked the new variables against `materialize-monitoring` v0.17.0 and the names, object shapes, defaults and sensitivity all match, with validation left upstream where it belongs; my only real question is the Azure zone story, where the module docs and the examples disagree about what a non-zonal AKS pool looks like.

## Summary

3 considerations
1 nit

bobbyiliev (approved):

azure/modules/monitoring/variables.tf
Alphadelta14: the note was kinda about non-labeled cases like on-prem just for completeness.

aws/examples/simple/main.tf
Alphadelta14: i think the comment makes sense saying to declare it as a variable or env first.

aws/modules/monitoring/variables.tf
Alphadelta14: please don't make me document every possible permutation LOL

PR #386, opened 2026-08-20
ory-stack: optional single-domain reverse proxy (spike)

bobbyiliev:
Adds an optional `single_domain_fqdn` toggle to the ory-stack module. When set, Hydra, Kratos and the UI sit behind one hostname, path-routed under /hydra, /kratos and /ui via a small nginx proxy that terminates TLS, instead of a hostname and LoadBalancer per service. Polis stays on its own hostname since the customer IdP has to reach it for the SAML POST. Default off, so existing deployments are unchanged.


kubernetes/modules/ory-stack/main.tf
```diff
@@ -29,30 +29,36 @@ locals {
...
+    var.selfservice_ui_url
+  )
   polis_external_url = local.wire_polis ? "https://${var.polis_fqdn}" : null
 
-  # Cookie domain shared across the Ory subdomains so flow/session cookies work
-  # across sibling FQDNs (Kratos, UI, Hydra). Defaults to the parent domain of
-  # kratos_fqdn (e.g. kratos.example.com -> example.com); when kratos_fqdn is a
-  # single label (no '.') we fall back to the value itself rather than erroring.
-  kratos_fqdn_parts = split(".", var.kratos_fqdn)
+  # Cookie domain shared across the Ory FQDNs. Defaults to the parent domain of
```
Alphadelta14: examples here would be nice. I know what it's doing but it's terribly complex to look at

kubernetes/modules/ory-stack/main.tf
```diff
@@ -368,15 +380,16 @@ module "ory_selfservice_ui" {
...
   kratos_browser_url = local.kratos_external_url
   hydra_admin_url    = local.hydra_admin_internal_url
 
-  tls_cert_secret_name = "ory-selfservice-ui-tls"
+  # Proxy terminates TLS in single-domain mode, so the UI serves plain HTTP.
+  tls_cert_secret_name = local.single_domain ? null : "ory-selfservice-ui-tls"
 
   # Only needed when Kratos/Hydra are served by the in-cluster self-signed CA.
-  trust_mounted_ca_cert = var.cert_issuer_signs_cluster_local
+  trust_mounted_ca_cert = !local.single_domain && var.cert_issuer_signs_cluster_local
```
Alphadelta14: it'd be nice to have a different variable name here. single_domain is not very obvious (like ory_protocol == http or somesuch)

kubernetes/modules/ory-stack/main.tf
```diff
@@ -613,3 +626,208 @@ resource "kubernetes_network_policy_v1" "ory_from_materialize_ingress" {
 
   depends_on = [kubernetes_namespace.ory]
 }
+
+# Single-domain reverse proxy (optional) -------------------------------------
```
Alphadelta14: i feel like we're inventing an ingress... I think Gateway APIs are available by default in most clusters at this point if we are interested in the internal routing side.

bobbyiliev: Hm yes true. I went with a self-contained proxy for the spike mainly because it behaves identically across GKE/EKS/AKS, with no dependency on a Gateway/ingress controller being installed and configured per cluster, and the path-prefix rewriting (stripping `/hydra`, `/kratos`, `/ui`) varies across ingress/Gateway implementations.

Open to suggestions though! If we'd rather standardize on Gateway API for internal routing, I'm happy to look into it.

Curious to hear what @jubrad thinks here too.

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Good call, added examples above the block showing the URLs in both modes

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Renamed it to `single_domain_enabled`. Kept the `public lb_addresses.single_domain` output key and `lb_overrides["single_domain"]` as-is as those are the module interface. But open to suggestions here!

kubernetes/modules/ory-stack/main.tf
Alphadelta14: we don't have prior art here yet, so we can keep it out of scope. definitely something to think of when introducing others. I may do it for loki so we get better experience here (I've done a lot of ingress/gateway management). Any time we are inventing something ourselves, we should be comparing build-it-ourselves vs adopt.

kubernetes/modules/ory-stack/main.tf
Alphadelta14: ah, i meant that single_domain doesn't quite imply needing to mount a ca cert IMHO, so having a cert-related variable makes more sense. I am be fine with !single_domain otherwise.

Alphadelta14 (approved):

jasonhernandez (approved):
Thanks- my next step will be to try and see this run in the context graph staging environment

kubernetes/modules/ory-stack/main.tf
```diff
@@ -613,3 +636,208 @@ resource "kubernetes_network_policy_v1" "ory_from_materialize_ingress" {
...
+          name = "pingap"
+          # Spike stopgap: public image. GA should pull a hardened, version-tracked
+          # pingap through the OEL registry proxy (cf. the cloud team's distroless build).
+          image             = "vicanso/pingap:0.12.1-full"
+          image_pull_policy = "IfNotPresent"
+          args              = ["-c", "/opt/pingap/conf"]
+
+          port {
+            name           = "https"
+            container_port = 8443
```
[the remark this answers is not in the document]
bobbyiliev: Good catch, this would have dropped all proxy traffic when Polis is off. It now opens 8443 whenever single-domain or Polis is on, and I limited 4444 and 4433 to per-service mode.

kubernetes/modules/ory-stack/main.tf
```diff
@@ -613,3 +636,208 @@ resource "kubernetes_network_policy_v1" "ory_from_materialize_ingress" {
...
+  ]
+}
+
+resource "kubernetes_service_v1" "single_domain_proxy_lb" {
+  count = local.single_domain_enabled ? 1 : 0
+
+  metadata {
+    name        = "${local.single_domain_proxy_name}-lb"
+    namespace   = var.namespace
+    annotations = merge(var.lb_annotations, try(var.lb_overrides["single_domain"].annotations, {}))
```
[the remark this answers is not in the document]
bobbyiliev: Agreed, single-domain mode was losing the firewall level restriction. I added single_domain to the allowed keys and the description so callers can set source_ranges on the proxy load balancer.

kubernetes/modules/ory-stack/main.tf
```diff
@@ -29,30 +29,46 @@ locals {
...
+  single_domain_enabled = var.single_domain_fqdn != null
+
+  # External URLs the browser (and Materialize, for OIDC issuer matching) sees.
+  # No trailing slash: OIDC issuer comparison downstream is exact-match.
+  hydra_external_url  = local.single_domain_enabled ? "https://${var.single_domain_fqdn}/hydra" : "https://${var.hydra_fqdn}"
+  kratos_external_url = local.single_domain_enabled ? "https://${var.single_domain_fqdn}/kratos" : "https://${var.kratos_fqdn}"
+  # /ui under the single domain, else the standalone UI FQDN, else whatever app
+  # hosts the flow pages (e.g. the console). Kratos/Hydra redirect the browser here.
+  ui_external_url = (
+    local.single_domain_enabled ? "https://${var.single_domain_fqdn}/ui" :
```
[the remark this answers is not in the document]
bobbyiliev: Fixed. It now checks deploy_selfservice_ui first, so the UI-off case uses selfservice_ui_url in both modes.

kubernetes/modules/ory-stack/main.tf
```diff
@@ -613,3 +636,208 @@ resource "kubernetes_network_policy_v1" "ory_from_materialize_ingress" {
...
+    proxy_set_headers = ["X-Forwarded-Proto:https"]
+    %{if var.deploy_selfservice_ui}
+    [upstreams.ui]
+    addrs = ["${local.single_domain_ui_upstream}"]
+    discovery = "dns"
+
+    [locations.ui]
+    path = "/ui"
+    upstream = "ui"
+    rewrite = "^/ui/(.*) /$1"
```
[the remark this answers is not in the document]
bobbyiliev: You are right. The UI serves its assets from the root, so it cannot run under a path prefix, so it now keeps its own hostname in single-domain mode and the proxy only fronts Hydra and Kratos. This also matters less longer term, since once the login flow is built into the console and we drop the Ory UI service the prefix problem goes away. I also fixed the bare prefix rewrites.

kubernetes/modules/ory-stack/main.tf
```diff
@@ -29,30 +29,46 @@ locals {
...
 
-  # Cookie domain shared across the Ory subdomains so flow/session cookies work
-  # across sibling FQDNs (Kratos, UI, Hydra). Defaults to the parent domain of
-  # kratos_fqdn (e.g. kratos.example.com -> example.com); when kratos_fqdn is a
-  # single label (no '.') we fall back to the value itself rather than erroring.
-  kratos_fqdn_parts = split(".", var.kratos_fqdn)
+  # Cookie domain shared across the Ory FQDNs. Defaults to the parent domain of
+  # the primary FQDN (single-domain host, else kratos_fqdn); single-label hosts
+  # fall back to the value itself.
+  cookie_primary_fqdn = local.single_domain_enabled ? var.single_domain_fqdn : var.kratos_fqdn
```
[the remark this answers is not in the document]
bobbyiliev: Makes sense. In single-domain mode the cookie now defaults to the single domain host itself instead of the parent, and I updated the description.

kubernetes/modules/ory-stack/main.tf
```diff
@@ -613,3 +636,208 @@ resource "kubernetes_network_policy_v1" "ory_from_materialize_ingress" {
...
+      }
+
+      spec {
+        node_selector = var.node_selector
+
+        container {
+          name = "pingap"
+          # Spike stopgap: public image. GA should pull a hardened, version-tracked
+          # pingap through the OEL registry proxy (cf. the cloud team's distroless build).
+          image             = "vicanso/pingap:0.12.1-full"
```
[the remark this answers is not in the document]
bobbyiliev: Thanks. I made the proxy image a variable, so we can point it at your hardened distroless build once it is published through the OEL registry proxy, and the proxy pod already uses the OEL pull secret so it will authenticate with the license key. For now I also added resource requests and limits and a small security context, and left non-root and read-only filesystem for the hardened image.

kubernetes/modules/ory-stack/main.tf
bobbyiliev: Good point. I pulled TLS termination into a tls_terminated_by_proxy local and use it at the Hydra and Kratos cert lines so they read as a TLS concern, and kept it derived from the toggle since single-domain is the only mode where the proxy terminates TLS for those services.

kubernetes/modules/ory-stack/main.tf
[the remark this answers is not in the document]
bobbyiliev: Quick correction after testing this end to end. Scoping to the single host actually broke the login, because the UI keeps its own sibling hostname in single-domain mode, so its host did not get the cookie and the flow looped. I reverted it back to the shared parent domain.

kubernetes/modules/ory-stack/main.tf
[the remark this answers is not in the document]
jubrad: @jasonhernandez can we just publish that image straight to materialize dockerhub?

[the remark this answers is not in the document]
jasonhernandez: sure!

jubrad (commented):
Round two: everything I raised is fixed or well answered — moving the UI back to its own hostname was the right call and the cookie-scope revert is correct — but the GCP example still documents the old `/ui` layout, which would send an operator into a broken DNS cutover.

## Summary

1 consideration

gcp/examples/enterprise/terraform.tfvars.example
```diff
@@ -19,6 +19,9 @@ k8s_apiserver_authorized_networks = [
 ory_hydra_fqdn             = "hydra.mz.example.com"
 ory_ui_fqdn                = "auth.mz.example.com"
 ory_kratos_fqdn            = "kratos.mz.example.com"
+# Optional: serve Hydra/Kratos/UI behind one hostname (/hydra, /kratos, /ui)
```
jubrad: **consider** — This still describes the pre-fix layout:

```
# Optional: serve Hydra/Kratos/UI behind one hostname (/hydra, /kratos, /ui)
# instead of the three above. Point one A record at lb_addresses.single_domain.
```

After the UI moved back to its own hostname, none of that holds: the proxy fronts Hydra and Kratos only, `ory_ui_fqdn` is still required, and it still needs its own A record pointed at `lb_addresses.ui`. An operator following this comment points one record, gets a UI hostname that doesn't resolve, and every login redirect dead-ends. That's the exact failure you hit yourself when the cookie was host-scoped, arrived at from the other direction.

Same edit needed in two more places that the module-side docs already got right: `gcp/examples/enterprise/variables.tf` (`ory_single_domain_fqdn` still says "Hydra, Kratos, and the UI ... (/hydra, /kratos, /ui) ... instead of the per-service hostnames above") and the `# When set, Hydra/Kratos/UI share one hostname behind a reverse proxy.` comment at `gcp/examples/enterprise/main.tf:659`.

bobbyiliev: Good catch, fixed all three. The tfvars example, the ory_single_domain_fqdn description, and the main.tf comment now say the proxy fronts Hydra and Kratos only, and that the UI keeps its own hostname and still needs its own A record pointed at lb_addresses.ui.

jubrad (approved):
Round three: the GCP example docs are fixed in all three places, which was the last thing I had blocking — this looks good to me now, with one optional precondition still open from last round.

PR #121, opened 2025-12-06
Autogenerate Helm and CRD Vars

jubrad:
```
- Add `materialize_spec_override` to materialize-instance module for arbitrary CRD fields
- Expose `helm_values_override` and `materialize_spec_override` in all cloud examples
- Change default authenticator_kind to Password for all deployments
- Add deepmerge provider for spec merging
- Add Python script to generate Terraform types from upstream CRD/Helm schemas
- Set up pyproject.toml with uv for dependency management
- Add GitHub Action to verify schema sync with upstream Materialize
- Update CONTRIBUTING.md with development setup and type generation docs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

```

Auto generating terraform variables for our CRD and Helm chart is something I've been thinking about for a while and this is a proof of concept for it. 

There are many benefits to having an object or set of vars that work with LSP or AI in order to help write code (as opposed to `object(any)`). And it's also extremely beneficial to be able to set any viable attribute for the helm chart or the materialize CR. 

PR has been updated to be a bit more reviewable. 
Commit 1 - adds generation script
Commit 2 - adds GH action and docs
Commit 3 - Adds the auto-generated files and adds deprecations
Commit 4 - Adds pytests (yes pytests) for basic validation of our input files to tf_vars
Commit 5 - Adds pytests that use kind and generated tf_vars to deploy a materialize operator via helm and a materialize instance

TODO:
- [ ] rebase / bring to newest MZ version
- [ ] Set static object types for kubernetes resources based on [type annotations ](https://github.com/MaterializeInc/materialize/pull/34775)
- [ ] [generate variables automatically in release ](https://github.com/MaterializeInc/materialize/pull/34488)

PR #387, opened 2026-08-20
CLO-213: Run self-managed module tests locally on kind

jubrad:
Linear issue: https://linear.app/materializeinc/issue/CLO-213/enable-self-managed-materialize-to-run-non-cloud-specific-module-tests

## Issue

Exercising the non-cloud-specific test surface (operator install, Materialize instance lifecycle, cert-manager/TLS wiring, SQL connectivity) previously required a full cloud deployment. This ports the kind-based flow from #121 to the rust test harness.

## Solution

Two commits:

1. **`kubernetes/examples/simple`** — a customer-facing example, symmetric with the cloud `examples/simple` roots, that deploys self-managed Materialize onto any existing cluster via a kubeconfig: cert-manager + self-signed issuer + the operator from the Helm chart (v1 CRD enabled) + a Materialize instance. The customer brings the metadata (PostgreSQL) and persist (S3-compatible) backends as connection-URL variables. Validated in the lint example matrix.
2. **A `kind` provider in the test harness** that deploys that example onto a local kind cluster, with the same lifecycle and ergonomics as the clouds:

```sh
cargo run -- run kind --owner "Your Name" --license-key-file key.txt
```

`init` creates a per-run kind cluster and deploys in-cluster PostgreSQL/RustFS test backends (`test/kind/backends.yaml`); `apply` deploys the example pointed at them via tfvars; `verify` runs the same checks as the clouds except the DNS ones, reaching SQL via `kubectl port-forward` to balancerd; `destroy` deletes the cluster. Iterating = keep the cluster, re-run `verify` (or `sync` + `apply`). Dev overrides (`--local-chart-path`, `--orchestratord-version`, `--environmentd-version`) work on kind, making it the cheapest way to smoke-test a local chart or image. Docs in `test/README.md`.

**CI decision:** `test-kind.yml` runs the full lifecycle but is `workflow_dispatch`/`workflow_call` only — deferred from the merge queue until proven reliable. It needs the `MATERIALIZE_LICENSE_KEY` secret and a `ubuntu-24.04-16core` runner (the stack's default resource requests don't fit a standard 4-CPU runner; the repo was granted access to the org's larger-runner group).

## Testing

- `cargo fmt/clippy -D warnings/deny`, `terraform fmt`, `tflint`, `terraform validate` all clean.
- Full lifecycle (`run kind`) passed end to end **locally** with a valid license key and **in CI** via workflow dispatch on this branch: instance UpToDate, all expected pods Running, `SELECT 1` through the port-forward, clean destroy.

## Gotchas for the reviewer

- The example sets `enable_network_policies = false` on materialize-instance: the module's egress allow-list only covers kube-system and the API server, which cuts environmentd off from backends the cluster hosts itself — recent kind versions *do* enforce NetworkPolicies, and the E2E run caught this. Enforcing-CNI users should enable it together with the chart's `networkPolicies` values (as the cloud operator modules do).
- The port-forward child's stdout must stay open for its lifetime (kubectl dies of EPIPE otherwise); the SQL check re-tunnels on every retry.
- Unlike #121, no public `kubernetes/modules/operator` module is added; the example installs the chart with a plain `helm_release`, mirroring the public install docs.
- `internal_load_balancer` / `grafana_allow_public_access` in `CommonTfVars` became `Option<bool>` (omitted for kind); the serialized tfvars for the clouds are unchanged.

### DIFF
--- a/.github/workflows/test-kind.yml
+++ b/.github/workflows/test-kind.yml
@@ -1,0 +1,60 @@
+name: Kind Tests
+
+# Self-managed Materialize on a local kind cluster. Needs no cloud
+# credentials, only the MATERIALIZE_LICENSE_KEY secret. Manually
+# dispatchable for now; wiring into the merge queue is deferred until the
+# workflow has proven reliable.
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+permissions:
+  contents: read # Required to checkout code
+
+jobs:
+  test-kind:
+    name: Kind Infrastructure Tests
+    # A larger runner: the stack's default resource requests (1 CPU / 4Gi
+    # for environmentd alone) do not fit the standard 4-CPU runner.
+    runs-on: ubuntu-24.04-16core
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4.0.1
+        with:
+          terraform_version: "1.15.8"
+
+      - name: Install kind
+        uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+        with:
+          install_only: true
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache Rust build artifacts
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            test/target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('test/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-
+
+      - name: Build tests
+        working-directory: test
+        run: cargo build
+
+      - name: Run Kind Tests
+        working-directory: test
+        env:
+          MATERIALIZE_LICENSE_KEY: ${{ secrets.MATERIALIZE_LICENSE_KEY }}
+        run: |
+          cargo run -- run --destroy-on-failure kind \
+            --owner "github-actions"
--- a/kubernetes/examples/simple/README.md
+++ b/kubernetes/examples/simple/README.md
@@ -12,7 +12,7 @@ This example deploys self-managed Materialize onto a Kubernetes cluster you alre
 
 - A Kubernetes cluster and a kubeconfig that reaches it
 - **Metadata backend**: a PostgreSQL database, passed as `metadata_backend_url`
-- **Persist backend**: an S3-compatible object store (S3, MinIO, GCS interop, ...), passed as `persist_backend_url`
+- **Persist backend**: an S3-compatible object store (S3, RustFS, GCS interop, ...), passed as `persist_backend_url`
 - A Materialize license key
 
 The instance's clusters schedule onto nodes labeled `materialize.cloud/swap=true` by default (see the chart's `clusterd.swapNodeSelector`), so label the nodes Materialize should run on accordingly.
--- a/kubernetes/examples/simple/README.md
+++ b/kubernetes/examples/simple/README.md
@@ -1,0 +1,38 @@
+# Example: Simple Materialize Deployment on an Existing Kubernetes Cluster
+
+This example deploys self-managed Materialize onto a Kubernetes cluster you already have -- on-prem, bare-metal, a local [kind](https://kind.sigs.k8s.io/) cluster, or any managed cluster -- using only a kubeconfig. Unlike the `aws/`, `azure/`, and `gcp/` examples, it provisions no cloud infrastructure.
+
+## What Gets Created
+
+- **cert-manager** and a **self-signed ClusterIssuer** for TLS between Materialize components
+- **Materialize operator** (`orchestratord`), installed from the [Helm chart](https://materializeinc.github.io/materialize/) into the `materialize` namespace, with the v1 CRD and its conversion webhook enabled
+- **Materialize instance** (`main`) in the `materialize-environment` namespace, with password authentication for `mz_system`
+
+## What You Bring
+
+- A Kubernetes cluster and a kubeconfig that reaches it
+- **Metadata backend**: a PostgreSQL database, passed as `metadata_backend_url`
+- **Persist backend**: an S3-compatible object store (S3, MinIO, GCS interop, ...), passed as `persist_backend_url`
+- A Materialize license key
+
+The instance's clusters schedule onto nodes labeled `materialize.cloud/swap=true` by default (see the chart's `clusterd.swapNodeSelector`), so label the nodes Materialize should run on accordingly.
+
+## Usage
+
+```sh
+terraform init
+terraform apply \
+  -var 'license_key=...' \
+  -var 'metadata_backend_url=postgres://user:pass@postgres.example.com:5432/materialize' \
+  -var 'persist_backend_url=s3://user:pass@bucket/prefix?endpoint=https%3A%2F%2Fs3.example.com&region=us-east-1'
+```
+
+Once applied, connect via the balancerd service (no load balancer is created):
+
+```sh
+kubectl port-forward -n materialize-environment "svc/$(terraform output -raw balancerd_service_name)" 6875:6875
+PGPASSWORD="$(terraform output -raw external_login_password_mz_system)" \
+  psql "sslmode=require host=127.0.0.1 port=6875 user=mz_system dbname=materialize"
+```
+
+This example is also what the test harness deploys onto a kind cluster for local/CI integration testing -- see `test/README.md`.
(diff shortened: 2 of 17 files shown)


jubrad:
Workflow run
https://github.com/MaterializeInc/materialize-terraform-self-managed/actions/runs/32443399352/job/96658407529

kubernetes/examples/simple/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  issuer_ref = {
+    name = module.self_signed_cluster_issuer.issuer_name
+    kind = "ClusterIssuer"
+  }
+
+  # The module's egress policies only allow kube-system and the API server,
+  # which cuts environmentd off from backends the cluster hosts itself or
+  # reaches on other ports. On enforcing CNIs, enable this together with the
+  # chart's networkPolicies values (see the cloud operator modules).
+  enable_network_policies = false
```
bobbyiliev: All four cloud examples set `enable_network_policies = true`. Could we keep it true here too by passing the chart's `networkPolicies` values in the helm_release above, the way `aws/modules/operator` does with `egress.cidrs = ["0.0.0.0/0"]`? Feels safer for an example customers copy from.

test/src/commands/init.rs
```diff
@@ -73,10 +76,30 @@ pub async fn phase_init(provider_args: &InitProvider) -> Result<PathBuf> {
 
         upload_tfvars_to_backend(&dest).await?;
 
+        // The cloud providers get their cluster from `terraform apply`; for
+        // kind the cluster is local infrastructure that terraform runs
+        // against, so it is created during init. After the tfvars are
+        // written, so a failed creation can still be cleaned up by `destroy`.
+        if provider == CloudProvider::Kind {
+            println!("\nCreating kind cluster {test_run_id}...");
+            create_kind_cluster(&test_run_id, &dest).await?;
```
bobbyiliev: Nice that a failed `terraform init` cleans up the cluster. If `create_kind_cluster` itself fails on the `kubectl apply` step though, the cluster is already up and `run --destroy-on-failure` never gets a dir back, so it would leak. Worth wrapping this call in the same cleanup?

test/kind/backends.yaml
```diff
@@ -0,0 +1,147 @@
...
+    matchLabels:
+      app: postgres
+  template:
+    metadata:
+      labels:
+        app: postgres
+    spec:
+      containers:
+        - name: postgres
+          image: postgres:16
```
bobbyiliev: Should these two get small resource requests? With none set they are BestEffort, so on a single node kind cluster they would be first to be evicted if Materialize pushes the node into memory pressure, which would show up as a confusing test failure.

kubernetes/examples/simple/main.tf
```diff
@@ -0,0 +1,79 @@
...
+
+module "self_signed_cluster_issuer" {
+  source = "../../modules/self-signed-cluster-issuer"
+
+  name_prefix = var.name_prefix
+
+  depends_on = [module.cert_manager]
+}
+
+resource "helm_release" "materialize_operator" {
```
bobbyiliev: Should this set a `timeout`? Helm defaults to 300s, and on a cold kind cluster the orchestratord pull plus the webhook cert can eat into that. The cert-manager module takes an `install_timeout` for the same reason.

kubernetes/examples/simple/versions.tf
```diff
@@ -0,0 +1,39 @@
+terraform {
+  required_version = ">= 1.10"
+
+  required_providers {
```
bobbyiliev: Should we add `/kubernetes/examples/simple` to `.github/dependabot.yml`? The other three `examples/simple` are listed there, and this one has a committed lock file too.

.github/workflows/test-kind.yml
```diff
@@ -0,0 +1,60 @@
...
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4.0.1
+        with:
+          terraform_version: "1.15.8"
```
bobbyiliev: Small one, main is on `1.15.9` now after #383, so this would land one patch behind the other workflows. Renovate keeps them all on the same exact version, so maybe worth matching on the rebase?

test/README.md
```diff
@@ -239,7 +267,11 @@ Same structure as AWS. Authenticates to GCP via Workload Identity Federation and
 
 Same structure as AWS. Authenticates to Azure via OIDC and to AWS via OIDC (for the S3 state backend). Skips if only AWS/GCP files changed.
 
-All three provider test workflows use `--destroy-on-failure` to ensure infrastructure is torn down even on test failure, and store Terraform state remotely in S3.
+### `test-kind.yml` -- Kind integration tests
+
+Runs the self-managed kind lifecycle (`cargo run -- run --destroy-on-failure kind`) on a plain GitHub-hosted runner. Needs no cloud credentials, only the `MATERIALIZE_LICENSE_KEY` secret. Currently `workflow_dispatch`/`workflow_call` only -- not wired into the merge queue until it has proven reliable there.
```
bobbyiliev: Small one, the workflow actually uses `ubuntu-24.04-16core` because the default requests do not fit a 4 CPU runner. Should this say a larger runner rather than a plain one?

test/src/commands/verify.rs
```diff
@@ -50,7 +55,11 @@ pub async fn phase_verify(dir: &Path) -> Result<()> {
         println!("\nVerifying the default DNS deployments were scaled down...");
```
bobbyiliev: Tiny thing, for kind this prints the header and then does nothing, because the list is empty. Worth an explicit skip message like the node-local-dns one just below?

kubernetes/examples/simple/README.md
```diff
@@ -0,0 +1,38 @@
+# Example: Simple Materialize Deployment on an Existing Kubernetes Cluster
```
bobbyiliev: Should we link this from the root README and `kubernetes/README.md`? The tree in the root README lists `examples/simple` for the three clouds but only `modules/` under `kubernetes/`, so this one is easy to miss.

kubernetes/examples/simple/README.md
```diff
@@ -0,0 +1,38 @@
...
+terraform apply \
+  -var 'license_key=...' \
+  -var 'metadata_backend_url=postgres://user:pass@postgres.example.com:5432/materialize' \
+  -var 'persist_backend_url=s3://user:pass@bucket/prefix?endpoint=https%3A%2F%2Fs3.example.com&region=us-east-1'
+```
+
+Once applied, connect via the balancerd service (no load balancer is created):
+
+```sh
+kubectl port-forward -n materialize-environment "svc/$(terraform output -raw balancerd_service_name)" 6875:6875
```
bobbyiliev: Since we output `materialize_instance_namespace`, should this use it rather than hardcoding the namespace? Just so the two cannot drift.

bobbyiliev (changes requested):
Nice work, mostly small things below, the network policy default in the example is the only one I would call a real question.

kubernetes/examples/simple/main.tf
jubrad: Done in 96f9880 — `enable_network_policies` is back to the module default and the helm_release now passes the chart's `networkPolicies` values (`internal` + `ingress`/`egress` `0.0.0.0/0`), matching the cloud operator modules. Validating on kind's enforcing CNI via the dispatch workflow (this exact combination is what the earlier E2E failure was about, so it gets a real run before merge).

test/src/commands/init.rs
jubrad: Good catch — fixed in 96f9880: `create_kind_cluster` failure now runs the same best-effort `kind delete cluster` as a failed `terraform init` (shared helper).

test/kind/backends.yaml
jubrad: Done in 96f9880 — both get `100m`/`256Mi` requests so they're Burstable rather than BestEffort.

kubernetes/examples/simple/main.tf
jubrad: Done in 96f9880 — `timeout = 600`.

kubernetes/examples/simple/versions.tf
jubrad: Done in 96f9880.

.github/workflows/test-kind.yml
jubrad: Done in 96f9880 — bumped to 1.15.9.

test/README.md
jubrad: Done in 96f9880 — now says larger (`ubuntu-24.04-16core`) runner and why.

test/src/commands/verify.rs
jubrad: Done in 96f9880 — prints an explicit skip when the provider has no deployments to check.

kubernetes/examples/simple/README.md
jubrad: Done in 96f9880 — added to the root README tree and linked from kubernetes/README.md's overview.

kubernetes/examples/simple/README.md
jubrad: Done in 96f9880 — uses `terraform output -raw materialize_instance_namespace`.

bobbyiliev (approved):
looks great!

PR #388, opened 2026-08-21
examples: lock the Polis SCIM endpoint to Okta's egress ranges

bobbyiliev:
Adds `scripts/update-okta-ip-ranges.sh`, which fetches Okta's published egress ranges for a given cell and writes `okta-scim-source-ranges.json`. The GCP enterprise example reads that into the Polis LoadBalancer's `loadBalancerSourceRanges` (via the module's `lb_overrides["polis"].source_ranges`), plus a new `ory_polis_source_ranges` var for your own VPC / VPN / tailnet CIDRs.

Ported from the equivalent setup in mz-context-graph. Opt-in and off by default: with no ranges file and an empty var, Polis stays unrestricted.

One caveat worth a look in review: this only fits deployments where browser access to Ory is internal (VPC / tailnet), because the browser hits Polis directly for the SAML ACS and OIDC-authorize hops, only Okta's SCIM push is server-to-server. For a fully public browser flow, Polis has to stay reachable from anywhere, so you'd leave this off. Wired on GCP for now; aws/azure can adopt the same `lb_overrides` pattern.

### DIFF
--- a/gcp/examples/enterprise/main.tf
+++ b/gcp/examples/enterprise/main.tf
@@ -193,6 +193,11 @@ locals {
 
   ory_namespace = "ory"
 
+  # Okta's SCIM egress ranges (generate with scripts/update-okta-ip-ranges.sh),
+  # plus any operator internal/VPN ranges, applied as the Polis LB allowlist.
+  okta_scim_source_ranges = fileexists("${path.module}/okta-scim-source-ranges.json") ? jsondecode(file("${path.module}/okta-scim-source-ranges.json")) : []
+  polis_lb_source_ranges  = concat(local.okta_scim_source_ranges, var.ory_polis_source_ranges)
+
   # cert-manager ClusterIssuer for browser-facing TLS. Defaults to the built-in
   # self-signed issuer; override via var.cert_issuer_ref to plug in a real one.
   cert_issuer = var.cert_issuer_ref != null ? var.cert_issuer_ref : {
@@ -682,6 +687,12 @@ module "ory" {
   # external NLB otherwise.
   lb_annotations = var.internal_load_balancer ? {
     "networking.gke.io/load-balancer-type" = "Internal"
+  } : {}
+
+  # Lock the Polis LB to Okta's SCIM egress ranges plus your internal ranges,
+  # when ory_polis_source_ranges is set or okta-scim-source-ranges.json exists.
+  lb_overrides = length(local.polis_lb_source_ranges) > 0 ? {
+    polis = { source_ranges = local.polis_lb_source_ranges }
   } : {}
 
   node_selector = local.generic_node_labels
--- a/gcp/examples/enterprise/variables.tf
+++ b/gcp/examples/enterprise/variables.tf
@@ -158,6 +158,13 @@ variable "polis_helm_values" {
   default     = {}
 }
 
+variable "ory_polis_source_ranges" {
+  description = "Extra CIDRs, besides Okta's SCIM egress ranges from okta-scim-source-ranges.json, allowed to reach the Polis LoadBalancer. Set these to your VPC / VPN / tailnet ranges so the browser SAML and OIDC hops still work, since the browser reaches Polis directly. Combined with the Okta ranges and applied as loadBalancerSourceRanges. Leave empty to keep Polis unrestricted. Only lock Polis down when browser access to Ory is internal; a fully public browser flow needs Polis reachable from anywhere."
+  type        = list(string)
+  default     = []
+  nullable    = false
+}
+
 variable "materialize_console_fqdn" {
   description = "External hostname for the Materialize console. Used to construct the OAuth2 redirect URI. Example: materialize.internal.example.com"
   type        = string
--- a/scripts/update-okta-ip-ranges.sh
+++ b/scripts/update-okta-ip-ranges.sh
@@ -1,0 +1,47 @@
+#!/bin/sh
+#
+# Writes Okta's published egress IP ranges for a given cell to
+# <example-dir>/okta-scim-source-ranges.json, so Terraform can apply them as
+# loadBalancerSourceRanges on the Polis LoadBalancer.
+#
+# Why: Okta pushes SCIM provisioning server-to-server from its cloud (no browser
+# in the loop), so the Polis LB can stay public but firewalled to Okta's egress
+# ranges plus your own internal ranges. See ory_polis_source_ranges in the
+# example and the README "Locking down the Polis endpoint" section.
+#
+# Okta publishes ranges per cell; find yours in the Okta admin console
+# (Settings). The Materialize tenant is us_cell_14 (the default here).
+#
+#   ./scripts/update-okta-ip-ranges.sh gcp/examples/enterprise us_cell_14
+#
+set -eu
+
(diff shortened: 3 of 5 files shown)


scripts/update-okta-ip-ranges.sh
```diff
@@ -0,0 +1,47 @@
...
+command -v curl >/dev/null 2>&1 || { echo "curl is required" >&2; exit 1; }
+command -v jq >/dev/null 2>&1 || { echo "jq is required" >&2; exit 1; }
+
+OUT="$DIR/okta-scim-source-ranges.json"
+
+curl -sSf "$OKTA_RANGES_URL" | jq --arg cell "$CELL" '
+    .[$cell].ip_ranges
+    | if . == null then error("cell \($cell) not found in ip_ranges.json") else . end
+    | sort
+' > "$OUT"
```
jubrad: **blocking** — `> "$OUT"` truncates the file before the pipeline runs, and `#!/bin/sh` + `set -eu` gives no `pipefail`, so the script's exit status is jq's. I reproduced both failure modes against a copy:

- Fetch fails (S3 blip, proxy, 403): jq gets empty input, emits nothing, exits 0. The script prints `Wrote  ranges for us_cell_14 to ...` and exits 0, and the previously-good `okta-scim-source-ranges.json` is now 0 bytes.
- Bad cell name: exits 5, but the file is likewise already truncated to 0 bytes.

Either way the next `terraform plan` fails on `jsondecode("")` with an opaque "unexpected end of JSON input", and the operator's allowlist is gone. Staging through a temp file fixes both:

```sh
TMP="$(mktemp)"
trap 'rm -f "$TMP"' EXIT
curl -sSf "$OKTA_RANGES_URL" > "$TMP"
jq --arg cell "$CELL" '...' "$TMP" > "$TMP.out"
mv "$TMP.out" "$OUT"
```

gcp/examples/enterprise/main.tf
```diff
@@ -709,6 +714,12 @@ module "ory" {
     "networking.gke.io/load-balancer-type" = "Internal"
   } : {}
 
+  # Lock the Polis LB to Okta's SCIM egress ranges plus your internal ranges,
+  # when ory_polis_source_ranges is set or okta-scim-source-ranges.json exists.
+  lb_overrides = length(local.polis_lb_source_ranges) > 0 ? {
```
jubrad: **consider** — `var.internal_load_balancer` defaults to `true` in this example, and `lb_annotations` applies `networking.gke.io/load-balancer-type: Internal` to every Ory LB including Polis. An internal passthrough NLB has a private VPC address, so Okta's SCIM push from its cloud cannot reach it at all — with the example's defaults, the Okta ranges this PR generates are dead weight on an LB Okta can never talk to.

So the README recipe implicitly requires `internal_load_balancer = false`, which also flips Hydra, Kratos and the selfservice UI to external. Could the README section say that explicitly? Alternatively, if the intent is internal LBs for everything else and a public Polis, the example could set `lb_overrides["polis"].annotations` to drop the Internal annotation for just that service — the module already merges overrides over `lb_annotations` with the override winning.

I may be misreading how you run this; if Polis is expected to stay internal, then the Okta half of the feature seems to have nothing to do, which is worth a sentence either way.

gcp/examples/enterprise/variables.tf
```diff
@@ -164,6 +164,13 @@ variable "polis_helm_values" {
   default     = {}
 }
 
+variable "ory_polis_source_ranges" {
+  description = "Extra CIDRs, besides Okta's SCIM egress ranges from okta-scim-source-ranges.json, allowed to reach the Polis LoadBalancer. Set these to your VPC / VPN / tailnet ranges so the browser SAML and OIDC hops still work, since the browser reaches Polis directly. Combined with the Okta ranges and applied as loadBalancerSourceRanges. Leave empty to keep Polis unrestricted. Only lock Polis down when browser access to Ory is internal; a fully public browser flow needs Polis reachable from anywhere."
```
jubrad: **consider** — "Set these to your VPC / VPN / tailnet ranges" is likely to mislead. `loadBalancerSourceRanges` becomes a cloud firewall rule matched against the source IP as GCP sees the packet — for a browser reaching an external LB over the internet that is the device's (or exit node's) public address, not its `100.64.0.0/10` tailnet address or a private `10.x` VPC CIDR. Someone who allowlists their tailnet CIDR verbatim will lock themselves out of the SAML ACS and OIDC-authorize hops and have a hard time working out why.

Worth saying "the public egress IP your browsers appear as (VPN/NAT egress, office range)", and keeping the private-CIDR advice for the internal-LB case only.

scripts/update-okta-ip-ranges.sh
```diff
@@ -0,0 +1,47 @@
...
+# <example-dir>/okta-scim-source-ranges.json, so Terraform can apply them as
+# loadBalancerSourceRanges on the Polis LoadBalancer.
+#
+# Why: Okta pushes SCIM provisioning server-to-server from its cloud (no browser
+# in the loop), so the Polis LB can stay public but firewalled to Okta's egress
+# ranges plus your own internal ranges. See ory_polis_source_ranges in the
+# example and the README "Locking down the Polis endpoint" section.
+#
+# Okta publishes ranges per cell; find yours in the Okta admin console
+# (Settings). The Materialize tenant is us_cell_14 (the default here).
```
jubrad: **consider** — This repo is public and these examples are what customers copy, so defaulting the cell to Materialize's own tenant is a sharp edge in both directions: it publishes which Okta cell we're on, and a customer who omits the argument silently gets someone else's ranges. That fails closed — their Okta SCIM push is blocked by an allowlist that looks plausible — which is exactly the kind of thing that burns an afternoon.

Since the README already writes it as `<your-okta-cell>`, I'd make the argument required and drop the default; the comment can still say where to find the cell in the Okta admin console.

jubrad (commented):
Sound approach, but the generator script needs a fix first — 1 blocker, 3 considerations.

scripts/update-okta-ip-ranges.sh
bobbyiliev: Good catch, fixed. Staged through a temp file, so a failed fetch or bad cell now leaves the existing allowlist untouched instead of a 0-byte file. Verified both modes.

gcp/examples/enterprise/main.tf
bobbyiliev: Right, with `internal_load_balancer = true` (default) Polis is internal and Okta can't reach it, so the ranges are inert. Made the README explicit that this needs `internal_load_balancer = false`. Kept it to docs since `lb_overrides` annotations only merge (can't drop the `Internal` one); making just Polis external would mean reworking the example's LB wiring. Happy to do that as a follow-up if you'd prefer.

gcp/examples/enterprise/variables.tf
bobbyiliev: Agreed, reworded the description and README to the browser's public egress IP (VPN/NAT egress), not a tailnet or private CIDR. We hit exactly this on the tailnet path, where the browser appears as the Cloud NAT IP.

scripts/update-okta-ip-ranges.sh
bobbyiliev: Done, cell is now required, the default is dropped, and our cell is out of the comments and usage.

jasonhernandez (approved):

PR #389, opened 2026-08-22
DEP-195 Support enabling certificates and auth for monitoring stack

Alphadelta14:
This upgrades materialize-monitoring to v0.18.x which has support for TLS between services. It also supports client authentication (which is used between services as well).
This does make use of cert-manager and would like to use other issuers (but we don't really have them built out yet), so it just uses its own for now.


bobbyiliev (commented):
A few small things, the upgrade note is the main one.

aws/modules/monitoring/variables.tf
```diff
@@ -533,6 +533,146 @@ variable "grafana_load_balancer" {
...
+
+    Issuing the material and using it are separate: `var.internal_tls` decides
+    how far the hops actually move off plaintext.
+  EOT
+  type        = bool
+  default     = true
+  nullable    = false
+}
+
+variable "internal_tls" {
```
bobbyiliev: This defaults to `authenticate`, and the description says a running stack should step through `present` first. Should we add a v12 note in the root README with that two step? #390 and #392 both add one, and this is the change that can drop telemetry on an upgrade.

aws/modules/monitoring/variables.tf
```diff
@@ -533,6 +533,146 @@ variable "grafana_load_balancer" {
...
+    settings point at. Null follows it — `authenticate` when certificates are on,
+    `off` when they are not — so turning certificates off needs one edit rather
+    than two. Naming a stage explicitly with certificates off is refused instead
+    of quietly downgraded.
+  EOT
+  type        = string
+  default     = null
+
+  validation {
+    condition     = var.internal_tls == null || contains(["off", "encrypt", "present", "authenticate"], coalesce(var.internal_tls, "off"))
```
bobbyiliev: Upstream already validates this exact list, and the note at line 214 in this file says validation stays upstream only. Worth dropping this block? Also the `coalesce` cannot do anything here, since the `== null` check just before it already short circuits.

aws/examples/simple/main.tf
```diff
@@ -653,7 +653,25 @@ module "monitoring" {
...
+  # deliberate, and the reason this differs from `module.materialize_instance`
+  # above: none of the components behind Grafana implements per-client
+  # authorization, so "signed by the CA we trust" is the whole authorization
+  # decision, and handing them the cluster's general-purpose issuer would reduce
+  # that to "has any certificate". Set `internal_issuer_ref` to share one
+  # knowingly.
+  #
+  # `issuer_ref` stays unset for a second reason: it is the *browser-facing*
+  # issuer, it only issues anything alongside `grafana_external_dns_names`, and
+  # these examples reach Grafana by port-forward rather than a public name.
```
bobbyiliev: Small thing, all six examples pass `grafana_load_balancer` unconditionally, so there is an NLB either way and setting `grafana_host` does give you a public name. Should this say the browser-facing cert is unset because no `grafana_host` is set by default, rather than because of port-forward?

aws/modules/monitoring/main.tf
```diff
@@ -695,6 +695,19 @@ module "monitoring" {
...
+  # secure here rather than on the monitoring module, because this repository's
+  # examples all install cert-manager and that module's other consumers may not.
+  certificates_enabled       = var.certificates_enabled
+  internal_tls               = coalesce(var.internal_tls, var.certificates_enabled ? "authenticate" : "off")
+  issuer_ref                 = var.issuer_ref
+  internal_issuer_ref        = var.internal_issuer_ref
+  grafana_external_dns_names = var.grafana_external_dns_names
+  certificate_duration       = var.certificate_duration
+  certificate_renew_before   = var.certificate_renew_before
+
```
bobbyiliev: Tiny thing, double blank line here. Same spot in the azure and gcp modules.

aws/modules/monitoring/variables.tf
Alphadelta14: i think it can mention it and it is pretty low risk because I don't think it's really landed without. The amount dropped is pretty minuscule (like a minute).

bobbyiliev (approved):
lgtm

jubrad (approved):
Clean, well-documented plumbing of the v0.18.x TLS surface, and I verified the variable names and semantics against upstream v0.18.1 — the main thing missing is a README upgrade note for what is a real contract change to existing deployments.

## Summary

1 must fix
2 considerations
1 nit

## Other

**blocking** — `aws/modules/monitoring/variables.tf:556` — This flips two defaults for every existing caller and there's no entry under README's `### Upgrade Notes`, which is where this repo has documented exactly this kind of change (see the v11.0.0 section).

Two distinct breakages on a plain `ref=` bump:

- A root that calls `aws/modules/monitoring` directly without cert-manager fails the apply on an unknown `cert-manager.io/v1` kind. The variable description says so, but nobody reads a variable description before bumping a tag.
- A root that *does* have cert-manager takes its running stack from plaintext straight to `authenticate` in one apply. Your own description for `internal_tls` says to apply `present` first and let it settle, and upstream is blunter ("on a busy pipeline is data loss rather than latency"). The default does the thing the docs warn against, and nothing tells the operator at bump time.

Worth folding into the same note: this bump also pulls in DEP-192, so NetworkPolicies are now rendered for every component and default to `enabled: true` — `templates/networkpolicies.yaml` doesn't exist at v0.17.0. That's a bigger blast radius than "TLS between services" and it isn't in the PR description either.

Same three lines in `gcp/` and `azure/`.

**consider** — `aws/examples/simple/main.tf:650` — "these examples reach Grafana by port-forward rather than a public name" isn't true any more — every one of the six examples sets `grafana_load_balancer` with `host = var.grafana_host`, and v11.0.0 was the release that made that so.

It matters because this sentence is the stated justification for leaving `issuer_ref` unset. The actual outcome of this PR's defaults on the examples is that every in-cluster hop gets mTLS while the admin login page is still served over plain HTTP through a load balancer, which is a defensible place to stop but a different statement than the one the comment makes. I'd rewrite it to say that (browser-facing TLS is out of scope here; the LB terminates nothing) rather than to claim there is no LB.

Same comment in all six example roots.

**consider** — `aws/modules/monitoring/main.tf:708` — The `grafana_service_port` TODO nine lines below now reads as an active hazard: "This becomes load-bearing with in-cluster TLS: Grafana stops serving plain HTTP, the Service port moves with it." With `internal_tls = authenticate` as the default, the next person here will believe the NLB is pointed at a port that no longer exists.

As far as I can tell it isn't — upstream's `internal_tls` docs say "Grafana's datasource verifies the backend but presents nothing", so Grafana is only a TLS client under these profiles and the Service stays 80 → 3000. Since this PR is the one that turns it on, it's the right place to correct that TODO (and the "which is DEP-195's work" line in the `grafana_load_balancer` description, which is now this PR).

**nit** — `aws/modules/monitoring/variables.tf:584` — The `coalesce` is dead — the left side of the `||` already short-circuits on null, so `var.internal_tls` is non-null by the time `contains` runs. `contains(["off", "encrypt", "present", "authenticate"], var.internal_tls)` says the same thing without implying null can reach it.

PR #392, opened 2026-08-24
DEP-201 Support account-namespaced buckets for mzmon

Alphadelta14:
This adds support for account-namespaced buckets in monitoring stack.
This is a breaking change that we would rather get out of the way sooner rather than later.
 

---


README.md
```diff
@@ -204,6 +209,14 @@ The AWS modules now require the `hashicorp/aws` provider `~> 6.0` (previously `~
...
 
+**Also in this release: `aws/modules/monitoring` moves its buckets to the S3 account regional namespace.** This is a separate change from the provider upgrade, and the destructive one. Read it before you apply.
+
+- **Both telemetry buckets are destroyed and recreated, and every Loki log and Thanos metric already in them is lost.** The buckets are now created in your account's regional namespace, which is house policy for new buckets: the name is reserved to your account, so no other account can take it and none can ever take it back. The name changes shape with it, from `<prefix>-mzmon-logs-<random>` to `<prefix>-mzmon-logs-<account>-<region>-an`, and a bucket's name forces a new resource. S3 offers no in-place migration between namespaces — AWS's own guidance is to create the new bucket and copy — so copy anything you need to keep first, or stay on `v11` until you have. Nothing else about the buckets changes: regional endpoints, ARNs, IAM resource patterns, and both backends' configuration are the same as before.
```
bobbyiliev: Since the buckets get destroyed with data still in them, would the destroy fail unless `bucket_force_destroy = true`? It defaults to false and the examples set it, so maybe worth a line here for anyone calling the module directly.

aws/modules/monitoring/main.tf
```diff
@@ -13,6 +13,14 @@
 # a legacy scrape config, collected metrics only, and ran a single Prometheus on
 # a ReadWriteOnce volume with 15 days of retention.
 #
+# UPGRADE NOTE — the telemetry buckets are REPLACED on the version that
```
bobbyiliev: Small one, could this name the version rather than "the version that introduced the account regional namespace"? The README note calls it v12, so it would be easier to tell whether you are past it.

bobbyiliev (commented):
Just a couple of small things.

bobbyiliev (dismissed):
lgtm

bobbyiliev (approved):

PR #395, opened 2026-08-25
DEP-201 Support account-namespaced buckets for mzmon

Alphadelta14:
This adds support for account-namespaced buckets in monitoring stack.
This is a breaking change that we would rather get out of the way sooner rather than later.

gh stacks closed https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/392 which was ready to go in...


bobbyiliev (approved):
lgtm

